PR #3, opened 2021-04-23
private join and compute crypto libs

stevenwarejones:
This adds the private join and compute dependencies and sets it up so bazel can build them.
I wasn't able to test the reencrypt and decrypt. They both throw an error `INVALID_ARGUMENT: ECGroup::CreateECPoint(string) - Could not decode point.` but I didn't want to spend much time debugging it. I think the implementation is probably correct but I'm not not sure how to properly test it.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3)


efoxepstein:
For `Decrypt` and `ReEncrypt`, you probably want a test that looks something like this:

```
absl::StatusOr<std::string> encrypted_value1 = cryptor1->Encrypt("some plaintext");
ASSERT_THAT(encrypted_value, IsOk());

absl::StatusOr<std::string> encrypted_value2 = cryptor2->Encrypt("some plaintext");
ASSERT_THAT(encrypted_value, IsOk());

absl::StatusOr<std::string> double_encrypted_value1 = cryptor2->ReEncrypt(*encrypted_value1);
ASSERT_THAT(double_encrypted_value1, IsOk());

absl::StatusOr<std::string> double_encrypted_value2 = cryptor1->ReEncrypt(*encrypted_value2);
ASSERT_THAT(double_encrypted_value2, IsOkAndHolds(*double_encrypted_value1));

EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value2));
EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value2));

EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value1));
EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value1));
```

build/com_google_private_join_and_compute/repo.bzl
```diff
@@ -0,0 +1,53 @@
...
+# limitations under the License.
+
+"""Repository rules/macros for private-join-and-compute library.
+
+See https://github.com/google/private-join-and-compute
+"""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def private_join_and_compute_repo(commit, sha256):
```
efoxepstein, line 22: This doesn't look like it's called... probably because it's accidentally not called in any of the other repos either. See https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/30/files for how we're fixing this in the other repo.

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+
+  absl::StatusOr<std::string> Decrypt(
+      absl::string_view encrypted_string) override;
+  absl::StatusOr<std::string> Encrypt(
+      absl::string_view plaintext) override;
+  absl::StatusOr<std::string> ReEncrypt(
+      absl::string_view encrypted_string) override;
+
+ private:
+  const std::unique_ptr<ECCommutativeCipher> local_ec_cipher_;
```
efoxepstein, line 50: Please add a thread annotation: `GUARDED_BY(mutex_)`... see https://abseil.io/docs/cpp/guides/synchronization#thread-annotations.

Please annotate the methods that acquire the Mutex with `LOCKS_EXCLUDED(mutex_)`.

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+};
+
+CryptorImpl::CryptorImpl(
+    std::unique_ptr<ECCommutativeCipher> local_ec_cipher)
+    : local_ec_cipher_(std::move(local_ec_cipher)) {}
+
+absl::StatusOr<std::string> CryptorImpl::Decrypt(
+    absl::string_view encrypted_string) {
+  absl::WriterMutexLock l(&mutex_);
+  ASSIGN_OR_RETURN(std::string decrypted_string,
```
efoxepstein, line 64: Simpler: `return local_ec_cipher_->Decrypt(encrypted_string));`

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+  return decrypted_string;
+}
+
+absl::StatusOr<std::string> CryptorImpl::Encrypt(
+    absl::string_view plaintext) {
+  absl::WriterMutexLock l(&mutex_);
+  ASSIGN_OR_RETURN(std::string encrypted_string,
+                   local_ec_cipher_->Encrypt(
+                       plaintext));
+  return encrypted_string;
```
efoxepstein, line 76: Simpler: `return local_ec_cipher_->Encrypt(plaintext));`

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+}
+
+absl::StatusOr<std::string> CryptorImpl::ReEncrypt(
+    absl::string_view encrypted_string) {
+  absl::WriterMutexLock l(&mutex_);
+  ASSIGN_OR_RETURN(std::string re_encrypted_string,
```
efoxepstein, line 82: Ditto

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+//We probably want to pass in a crypto key in the future. This is just a placeholder.
+absl::StatusOr<std::unique_ptr<Cryptor>> CreateCryptorWithNewKey(void) {
+  ASSIGN_OR_RETURN(
+      auto local_ec_cipher,
+      ECCommutativeCipher::CreateWithNewKey(
+          NID_X9_62_prime256v1, ECCommutativeCipher::HashType::SHA256));
+  std::unique_ptr<Cryptor> result =
+      absl::make_unique<CryptorImpl>(
+          std::move(local_ec_cipher));
+  return {std::move(result)};
```
efoxepstein, line 99: `return absl::make_unique<..>(...)` should likely work

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
```diff
@@ -0,0 +1,113 @@
...
+
+absl::StatusOr<std::unique_ptr<Cryptor>> CreateCryptorFromKey(absl::string_view key_bytes) {
+  ASSIGN_OR_RETURN(
+      auto local_ec_cipher,
+      ECCommutativeCipher::CreateFromKey(
+          NID_X9_62_prime256v1, key_bytes, ECCommutativeCipher::HashType::SHA256));
+  std::unique_ptr<Cryptor> result =
+      absl::make_unique<CryptorImpl>(
+          std::move(local_ec_cipher));
+  return {std::move(result)};
```
efoxepstein, line 110: `return absl::make_unique<..>(...)` should likely work

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h
```diff
@@ -0,0 +1,57 @@
...
+class Cryptor {
+ public:
+  virtual ~Cryptor() = default;
+
+  Cryptor(Cryptor&& other) = delete;
+  Cryptor& operator=(Cryptor&& other) = delete;
+  Cryptor(const Cryptor&) = delete;
+  Cryptor& operator=(const Cryptor&) = delete;
+
+  virtual absl::StatusOr<std::string> Decrypt(
```
efoxepstein, line 40: What do you think about supporting a batch interface here to avoid needing to re-acquire the Mutex between each low-level operation? Maybe `absl::StatusOr<std::vector<std::string>> Whatevercrypt(absl::Span<const absl::string_view> plaintexts_or_ciphertexts)`?

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "absl/status/status.h"
+#include "absl/status/statusor.h"
+#include "absl/strings/string_view.h"
+#include "absl/types/span.h"
+
+#include "crypto/ec_commutative_cipher.h"
+
+namespace wfanet::panelmatch::common::crypto {
+using ::private_join_and_compute::ECCommutativeCipher;
```
efoxepstein, line 29: This looks unused in this header

efoxepstein (changes requested):

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc
stevenwarejones: When, in general, should I be using ASSIGN_OR_RETURN vs just returning? Decrypt/Encrypt/ReEncrypt can throw errors.

efoxepstein: This logic, when unpacked and in pseudocode, is basically:

```
if (there is an error) { return the statusor; }
unpack the value from the statusor
build and return a new statusor from the value
```

The rule is this: if you're returning something of type `absl::StatusOr<Foo>` and you have an `absl::StatusOr<Foo>` and you don't care whether or not it's an error or contains a Foo, just return what you have, don't bother unpacking it.

stevenwarejones:
> For `Decrypt` and `ReEncrypt`, you probably want a test that looks something like this:
> 
> ```
> absl::StatusOr<std::string> encrypted_value1 = cryptor1->Encrypt("some plaintext");
> ASSERT_THAT(encrypted_value, IsOk());
> 
> absl::StatusOr<std::string> encrypted_value2 = cryptor2->Encrypt("some plaintext");
> ASSERT_THAT(encrypted_value, IsOk());
> 
> absl::StatusOr<std::string> double_encrypted_value1 = cryptor2->ReEncrypt(*encrypted_value1);
> ASSERT_THAT(double_encrypted_value1, IsOk());
> 
> absl::StatusOr<std::string> double_encrypted_value2 = cryptor1->ReEncrypt(*encrypted_value2);
> ASSERT_THAT(double_encrypted_value2, IsOkAndHolds(*double_encrypted_value1));
> 
> EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value2));
> EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value2));
> 
> EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value1));
> EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value1));
> ```

What do you think about using something like cucumber-cpp for the tests? https://github.com/cucumber/cucumber-cpp/blob/master/examples/Calc/features/addition.feature ... I'm always a fan of human readable tests.

src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h
stevenwarejones: That sounds great and makes sense. I'll add that in.

stevenwarejones: @efoxepstein is there a standard length of a plaintext?

efoxepstein: In some sense it doesn't matter -- what happens under the hood is:

1. Any string is hashed to point on an elliptic curve
2. The point itself is encrypted
3. The hash itself isn't reversible (but since hashing is deterministic, we can [but don't actually need to] build lookup tables)

We expect the plaintexts to be email addresses, user ids, etc.

efoxepstein:
> > For `Decrypt` and `ReEncrypt`, you probably want a test that looks something like this:
> > ```
> > absl::StatusOr<std::string> encrypted_value1 = cryptor1->Encrypt("some plaintext");
> > ASSERT_THAT(encrypted_value, IsOk());
> > 
> > absl::StatusOr<std::string> encrypted_value2 = cryptor2->Encrypt("some plaintext");
> > ASSERT_THAT(encrypted_value, IsOk());
> > 
> > absl::StatusOr<std::string> double_encrypted_value1 = cryptor2->ReEncrypt(*encrypted_value1);
> > ASSERT_THAT(double_encrypted_value1, IsOk());
> > 
> > absl::StatusOr<std::string> double_encrypted_value2 = cryptor1->ReEncrypt(*encrypted_value2);
> > ASSERT_THAT(double_encrypted_value2, IsOkAndHolds(*double_encrypted_value1));
> > 
> > EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value2));
> > EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value2));
> > 
> > EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value1));
> > EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value1));
> > ```
> 
> What do you think about using something like cucumber-cpp for the tests? https://github.com/cucumber/cucumber-cpp/blob/master/examples/Calc/features/addition.feature ... I'm always a fan of human readable tests.

For me, I prioritize consistency and explicitness in unit tests. Magic is kind of scary. I _really_ don't want regexps in my test cases (unless that's part of the API of the library I'm testing)!

The tech infra workstream aligned on using googletest in C++. So, I would strongly urge you to use googletest or bring this issue up with the tech infra workstream.

(If you haven't seen this talk, I'd recommend it: https://www.youtube.com/watch?v=u5senBJUkPc)

stevenwarejones:
> > > For `Decrypt` and `ReEncrypt`, you probably want a test that looks something like this:
> > > ```
> > > absl::StatusOr<std::string> encrypted_value1 = cryptor1->Encrypt("some plaintext");
> > > ASSERT_THAT(encrypted_value, IsOk());
> > > 
> > > absl::StatusOr<std::string> encrypted_value2 = cryptor2->Encrypt("some plaintext");
> > > ASSERT_THAT(encrypted_value, IsOk());
> > > 
> > > absl::StatusOr<std::string> double_encrypted_value1 = cryptor2->ReEncrypt(*encrypted_value1);
> > > ASSERT_THAT(double_encrypted_value1, IsOk());
> > > 
> > > absl::StatusOr<std::string> double_encrypted_value2 = cryptor1->ReEncrypt(*encrypted_value2);
> > > ASSERT_THAT(double_encrypted_value2, IsOkAndHolds(*double_encrypted_value1));
> > > 
> > > EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value2));
> > > EXPECT_THAT(cryptor1->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value2));
> > > 
> > > EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value1), IsOkAndHolds(*encrypted_value1));
> > > EXPECT_THAT(cryptor2->Decrypt(*double_encrypted_value2), IsOkAndHolds(*encrypted_value1));
> > > ```
> > 
> > 
> > What do you think about using something like cucumber-cpp for the tests? https://github.com/cucumber/cucumber-cpp/blob/master/examples/Calc/features/addition.feature ... I'm always a fan of human readable tests.
> 
> For me, I prioritize consistency and explicitness in unit tests. Magic is kind of scary. I _really_ don't want regexps in my test cases (unless that's part of the API of the library I'm testing)!
> 
> The tech infra workstream aligned on using googletest in C++. So, I would strongly urge you to use googletest or bring this issue up with the tech infra workstream.
> 
> (If you haven't seen this talk, I'd recommend it: https://www.youtube.com/watch?v=u5senBJUkPc)

I think thats fine. I kind of like tests that require self-documentation. I can just add further documentation instead in the future.

efoxepstein (changes requested):
> Reviewed 6 of 9 files at r1, 2 of 5 files at r2, 6 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3)* status: 13 of 14 files reviewed, 14 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[keyfile.key, line 0 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZE9KZcCmG9hlOjR-CI:-MZE9KZd5CQWEIqkStQk:b6f6146) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/keyfile.key#L0)):*
Was this intended to be added?
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 27 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZE9fwb99LnVm3Bps32:-MZE9fwb99LnVm3Bps33:b8all51) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L27)):*
> ```objc
> 
> namespace wfanet::panelmatch::common::crypto {
> enum Action {
> ```

Please use an "enum class" instead of just an enum to improve scoping.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 29 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZE9jWX5nhWVHJ0qrRL:-MZE9jWX5nhWVHJ0qrRM:bm7d3mq) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L29)):*
> ```objc
> enum Action {
>   Encrypt,
>   ReEncrypt,
> ```

go/cstyle#Enumerator_Names -- please name like "kCamelCase" -- start with letter "k".
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZE9wlgDewsPkQMlr-7:-MZE9wlgDewsPkQMlr-8:botwwzu) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L48)):*
> ```objc
>   virtual absl::StatusOr<std::string> ReEncrypt(
>         absl::string_view encrypted_string) = 0;
>   virtual absl::StatusOr<std::vector<std::string>> BatchProcess(
> ```

Totally up to you, but it wouldn't be unreasonable for this to be the ONLY thing in the public API. It would be simpler to test and maintain. (And it's likely the only thing to be actually used.)
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 49 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZEA69LEcynjWtYiIYR:-MZEA69LEcynjWtYiIYS:bm8n6jg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L49)):*
> ```objc
>         absl::string_view encrypted_string) = 0;
>   virtual absl::StatusOr<std::vector<std::string>> BatchProcess(
>         std::vector<std::string> plaintexts_or_ciphertexts,
> ```

https://abseil.io/tips/93 recommends using absl::Span here -- it's super efficient and makes the API mostly agnostic to the type of container.

The way you have it now will require std::moveing a vector or making a copy -- see https://abseil.io/tips/148 for why an overload set is preferred to pass-by-value for containers. But the overload set isn't actually necessary here because we don't need a parameter sink -- there shouldn't be any need to mutate the input.

A question to revisit later is if we want an absl::Span<const std::string> or absl::Span<const absl::string_view> here. OR we might want a `const RepeatedPtrField&` if the data is coming right from a protocol buffer.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 85 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZEC7M1CQlZBi3fB7Kj:-MZEC7M1CQlZBi3fB7Kk:b-9p6jwi) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L85)):*
> ```cpp
>     std::vector<std::string> plaintexts_or_ciphertexts,
>     const Action action) {
>   std::vector<std::string> results;
> ```

Maybe call `results.reserve(plaintexts_or_ciphertexts.size());` here to avoid multiple allocations in the loop.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 91 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZECMOLD-d_O605hQXv:-MZECMOLD-d_O605hQXw:bbi59kr) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L91)):*
> ```cpp
>       case Action::Encrypt: {
>         ASSIGN_OR_RETURN(auto encrypted_string,
>                          Encrypt(text));
> ```

We'll want to avoid reacquiring the mutex each iteration of the loop -- instead, we should be able to grab it once right before the loop.
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 21 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZECf4q9rgj1YpphirZ:-MZECf4q9rgj1Ypphir_:b26iuhd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L21)):*
> ```cpp
> using ::wfanet::panelmatch::common::crypto::CreateCryptorWithNewKey;
> using ::wfanet::panelmatch::common::crypto::CreateCryptorFromKey;
> using ::wfanet::panelmatch::common::crypto::Action;
> ```

https://abseil.io/tips/119 recommends putting using-declarations in the innermost namespace ("Declare namespace aliases and using-declarations inside the innermost namespace, whether named or anonymous")
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 28 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZECoOj6Zr12LdV17Qz:-MZECoOj6Zr12LdV17R-:b-kb13eo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L28)):*
> ```cpp
> namespace wfa::panelmatch {
> namespace {
>   TEST(PrivateJoinAndComputeTest, EncryptReEncryptDecrypt) {
> ```

Please run clang-format on this -- https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/docs/code-style.md#other-languages

(It looks like everything should be dedented once.)
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 30 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZEDC0XASr2EDSXAbPv:-MZEDC0XASr2EDSXAbPw:b6k3pr7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4c6de03fe57249ee4ecb3f66aad079f0cb62a112/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L30)):*
> ```cpp
>   TEST(PrivateJoinAndComputeTest, EncryptReEncryptDecrypt) {
>     std::unique_ptr<Cryptor> cryptor1 = std::move(CreateCryptorFromKey("9ias9fi0s").value());
>     std::unique_ptr<Cryptor> cryptor2 = std::move(CreateCryptorFromKey("asdfasdfs").value());
> ```

If CreateCryptorFromKey fails, bad things will happen. This is safer:

ASSERT_OK_AND_ASSIGN(std::unique_ptr<Cryptor> cryptor1, CreateCryptorFromKey("9ias9fi0s"));

efoxepstein (changes requested):
> Reviewed 1 of 1 files at r4, 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3)* status: all files reviewed, 16 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/cc/wfanet/panelmatch/common/macros.h, line 41 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZF9UXuAPBa5ehHMxac:-MZF9UXv9JuB7jVTKY7r:blu11w5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/707bad2581d17f33a5e87d40ea915008802deafd/src/main/cc/wfanet/panelmatch/common/macros.h#L41)):*
> ```
> "
> ```

Extra quotes?
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 79 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZF9xDvDnlx9Xtsfagd:-MZF9xDvDnlx9Xtsfage:b-7iepcd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/707bad2581d17f33a5e87d40ea915008802deafd/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L79)):*
> ```cpp
>     std::string plaintext4 = "some plaintext4";
>     std::vector <std::string> plaintext_v {
>           plaintext0,
> ```

Maybe just inline the strings into the vector -- it doesn't seem like they're referred to directly.

stevenwarejones:
I did everything, I think, that you requested except converting the input/output of batchProcess to absl::span from vector. I had a couple questions. I was also thinking that if I'm going to convert it to `const RepeatedPtrField&` I should just do that anyways. Is that okay to do separately?

efoxepstein (changes requested):
Yeah, that's fine to do separately. Let's make it a "const std::vector<std::string>&" for now, though, to avoid the additional copy.

> Reviewed 7 of 9 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-:-MZJ7Z6AA_LWyN9mY65c:bjeqvzo)* status: 13 of 15 files reviewed, 12 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 27 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJBURdC1RnXTzTtWJq:-MZJBUReBThPV7m4jUON:b-qq7no2) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L27)):*
https://abseil.io/tips/130 recommends shallow namespace hierarchies.

> Don’t nest deeply: a single top-level namespace per project gets the same result without long/complicated names, with less exposure to accidents, without causing surprise for new engineers, and without the need to build any tooling.

Our other C++ code does this wrong, but there's an effort to fix it. I'll try to dig up the doc with the current recommended namespace naming guidelines for this codebase. I'm fine with submitting as is, as long as we refactor soon.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 41 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJ8SwSAH4PmBP3OiX-:-MZJ8SwSAH4PmBP3OiX0:btc0d5a) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L41)):*
> ```objc
>   virtual absl::StatusOr<std::vector<std::string>> BatchProcess(
>       std::vector<std::string> plaintexts_or_ciphertexts,
>       const Action action) = 0;
> ```

This "const" actually doesn't do anything -- it's actually quite misleading: https://abseil.io/tips/109
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 47 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJ84wy68AgGEWuu39-:-MZJ84wy68AgGEWuu390:b-nrd4kj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L47)):*
> ```objc
> 
>  private:
>   absl::StatusOr<std::string> Decrypt(absl::string_view encrypted_string);
> ```

I think you don't really need these here in the header -- I'd just delete.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 95 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJ8h6xByOpJaIZy_qJ:-MZJ8h6xByOpJaIZy_qK:b4wxfbf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L95)):*
> ```cpp
>       case Action::kDecrypt: {
>         ASSIGN_OR_RETURN(auto decrypted_string, Decrypt(text));
>         results.push_back(decrypted_string);
> ```

This is a completely optional comment, do with it what you will. Since each case is appending to results, you can pull this out pretty easily to make it clear that the only difference in cases is which function:

```
for (const std::string& text : plaintexts_or_ciphertexts) {
  std::string result;
  switch (action) {
    case Action::kFoo: {
      ASSIGN_OR_RETURN(result, ...);
      break;
     }
  }
  results.push_back(std::move(result));
}
```
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 29 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJ9a-79Qw1M9PepYLt:-MZJ9a-87hKv6ihMsP7h:b-foexxr) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L29)):*
> ```cpp
> #include "src/test/cc/testutil/status_macros.h"
> 
> using ::private_join_and_compute::ECCommutativeCipher;
> ```

Ditto -- https://abseil.io/tips/119 recommends putting using-declarations in the innermost namespace ("Declare namespace aliases and using-declarations inside the innermost namespace, whether named or anonymous")
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 39 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJA6hG7LKctTYrMr2c:-MZJA6hG7LKctTYrMr2d:b-ej28s) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L39)):*
> ```cpp
> using ::wfanet::panelmatch::common::crypto::CreateCryptorWithNewKey;
> using ::wfanet::panelmatch::common::crypto::Cryptor;
> TEST(PrivateJoinAndComputeTest, EncryptReEncryptDecrypt) {
> ```

Maybe another simple test case to add is that the encryption functions are NOT the identity function. Since the inputs and outputs are basically the same format, that would guard against any bugs for accidentally returning the input or something.
___
*[src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc, line 40 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJ9mGvBW7uPBmFSY5s:-MZJ9mGvBW7uPBmFSY5t:b6m5lll) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/58777aceddc048f923532c8a0eea48be2e075a0d/src/test/cc/main/wfanet/panelmatch/common/crypto/cryptor_test.cc#L40)):*
> ```
> 9ias9fi0
> ```
 
It's not super clear to the reader if these strings are just you mashing on the keyboard or if there's some structure to them -- to make it more self-documenting, maybe use strings like "some-arbitrary-key-bytes-1" or something?

efoxepstein (approved):
> Reviewed 2 of 9 files at r6, 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 83 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/3#-MZJpOFa0PprheC-DeHy:-MZJpOFa0PprheC-DeHz:bmlpxnd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3736f7ba1716f6323cde4aa26cf237c121008805/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L83)):*
> ```cpp
>   for (auto& text : plaintexts_or_ciphertexts) {
>     std::string result;
>     switch (action) {
> ```

OK last comment here, I promise! If you pull out the switch statement to a helper function, things get a lot simpler:

```
absl::StatusOr<std::string> PerformAction(Action action, absl::string_view text) {
  switch (action) {
    case Action::kEncrypt: return Encrypt(text);
    case Action::kReEncrypt: return ReEncrypt(text);
    case Action::kDecrypt: return Decrypt(text);
  }
  return absl::InvalidArgumentError("Unknown action");
}
```

And then here:

```
for (auto& text : plaintexts_or_ciphertexts) {
  ASSIGN_OR_RETURN(std::string result, PerformAction(action, text));
  results.push_back(std::move(result));
}
```

I like this because (1) it gets rid of breaks and blocks in the switch, (2) it's easier to reason about logic outside of loops, (3) it's easier to see what the loop is doing. WDYT?

PR #6, opened 2021-04-30
JNI commutative encryption

stevenwarejones:
* JNI/Swig for Commutative Encrypt/Decrypt/ReEncrypt

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)


src/main/kotlin/wfanet/panelmatch/common/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+
+package(default_visibility = [
+    "//src/main/kotlin/wfanet/panelmatch:__subpackages__",
+    "//src/test/kotlin/wfanet/panelmatch:__subpackages__",
+])
+
+kt_jvm_library(
+    name = "common",
+    srcs = glob(["*.kt"]),
+    runtime_deps = [
```
stevenwarejones, line 11: remove

src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc
```diff
@@ -0,0 +1,153 @@
...
+  auto encrypted_texts1 = (*encrypted_response1).encrypted_texts();
+
+  ApplyCommutativeEncryptionRequest encrypt_request2;
+  encrypt_request2.set_encryption_key(random_key_2);
+  encrypt_request2.mutable_plaintexts()->CopyFrom(plaintext_batch);
+  auto encrypted_response2 = ApplyCommutativeEncryption(encrypt_request2);
+  ASSERT_THAT(encrypted_response2, IsOk());
+  auto encrypted_texts2 = (*encrypted_response2).encrypted_texts();
+
+  assert_rpf_not_equal(encrypted_texts2, encrypted_texts1);
```
stevenwarejones, line 96: there has to be a cleaner way to do this @efoxepstein but I thought I would just ask you in review. What's the easiest way to compare to values of two RepeatedPtrFields<string>?

efoxepstein (changes requested):
> Reviewed 40 of 48 files at r1, 3 of 8 files at r2, 7 of 7 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 14 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

___
*[build/wfa/repositories.bzl, line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZb1W20hyrnSnhCLYK:-MZZb1W3BcQkTVTrpQVb:bo33qqi) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/build/wfa/repositories.bzl#L1)):*
> ```Starlark
> # Copyright 2021 The Cross-Media Measurement Authors
> ```

For this and all the other files in `build`, are these all copied directly from `cross-media-measurement`? Are we able to just take a dependency on `cross-media-measurement` (for now) and refer to these files in that repo?
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZbfO4EqRbqqzDqNUu:-MZZbfO4EqRbqqzDqNUv:bbwlbjo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L30)):*
> ```objc
> 
> namespace wfanet::panelmatch::common::crypto {
> using google::protobuf::RepeatedPtrField;
> ```

A using-declaration in a header file means that we're exporting the symbol as part of the public API of the library.

Since we don't expect anyone to write "wfanet::panelmatch::common::crypto::RepeatedPtrField", it's probably best to avoid the using-declaration and just spell out the full type name.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZbu_w-U71ukcCRNcY:-MZZbu_w-U71ukcCRNcZ:b-uq0c83) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L54)):*
> ```objc
> 
>  private:
>   absl::StatusOr<std::string> Decrypt(absl::string_view encrypted_string);
> ```

These can be removed, I think. There's no real point to private methods in an interface, is there? Nothing could call them.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZcH7c-mUwbW1xjr7Z:-MZZcH7dB_9ZtvY2xe7v:b-augwer) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L118)):*
> ```cpp
>     std::string result;
>     switch (action) {
>       case Action::kEncrypt: {
> ```

Rather than duplicate this with above, let's just pull out the switch to a helper function.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.h, line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZcRCz4j9ym52zhFvI:-MZZcRCz4j9ym52zhFvJ:bfprp8e) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.h#L26)):*
> ```objc
> 
> template <typename T>
> absl::Status ParseRequestFromString(T& request_proto,
> ```

I'm guessing this is copied from the other repository, but while you're here, we should fix it up.

The style guide prefers input parameters before outputs: "When ordering function parameters, put all input-only parameters before any output parameters." (https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs)
___
*[src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.cc, line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZd38R8m9oaEv8ZBWO:-MZZd38S04SCl7cJ-iww:b-ze0b73) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.cc#L17)):*
> ```cpp
> #include "wfanet/panelmatch/common/crypto/encryption_utility_helper.h"
> 
> namespace wfanet::panelmatch::common::crypto {
> ```

Delete this file?
___
*[src/main/cc/wfanet/panelmatch/common/crypto/started_thread_cpu_timer.h, line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZd90i6WcQ4-WO1vct:-MZZd90i6WcQ4-WO1vcu:bfd8pwm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/started_thread_cpu_timer.h#L26)):*
> ```objc
> 
> // Gets the cpu duration of current thread.
> absl::Duration GetCurrentThreadCpuDuration() {
> ```

Can we just take a dependency on some other repository instead of copying this stuff over? Should be better to maintain. Eventually, we'll pull out all the common stuff to a utility repo.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.h, line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdMSu3qFvq-9sdjHX:-MZZdMSu3qFvq-9sdjHY:b-al08mk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.h#L23)):*
> ```objc
> namespace wfanet::panelmatch::protocol::crypto {
> 
> using ::wfanet::panelmatch::protocol::protobuf::
> ```

See previous comment on using-declarations in headers... this one could be considered OK, but since each of these things is only referenced once, doesn't really seem worthwhile to add the using-declarations.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdmpT8dQDKBUtIpvd:-MZZdmpT8dQDKBUtIpve:b-xgocpj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L51)):*
> ```cpp
>                          "Failed to create the protocol cipher");
>   ASSIGN_OR_RETURN(
>       auto encrypted_texts,
> ```

I would guess that you could write `ASSIGN_OR_RETURN(*response.mutable_whatever(), cryptor->Whatever())` -- does that work?
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_wrapper.cc, line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZeQAz3-HRhtwYui-N:-MZZeQAz3-HRhtwYui-O:b-rz9yg5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_wrapper.cc#L45)):*
> ```cpp
> absl::StatusOr<std::string> ApplyCommutativeEncryptionWrapper(
>     const std::string& serialized_request) {
>   ApplyCommutativeEncryptionRequest request_proto;
> ```

These feel like they're begging to be templated:

```
template <typename RequestT, typename ResponseT>
absl::StatusOr<std::string> SomeGoodName(const std::string& serialized_request, absl::FunctionRef<ResponseT(RequestT)> fn) {
  RequestT request_proto;
  RETURN_IF_ERROR(ParseRequestFromString(request_proto, serialized_request));
  ASSIGN_OR_RETURN(ResponseT response, fn(request_proto));
  return response.SerializeAsString(); 
}

absl::StatusOr<std::string> ReApplyCommutativeEncryptionWrapper(
    const std::string& serialized_request) {
  return SomeGoodName(serialized_request, ReApplyCommutativeEncryption);
}
```

I dunno, what do you think?
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZa_8Sen3uGotJs4W2-r1-96:-MZZgZql2A9zMipSlm4q:b-a0deuh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/0bcdccd97a851ab5463a63c6c44456aa32ef257c/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L96)):*

there has to be a cleaner way to do this @efoxepstein but I thought I would just ask you in review. What's the easiest way to compare to values of two RepeatedPtrFields<string>?
</blockquote></details>

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/test/cc/testutil/matchers.h#L48

```
ASSERT_THAT(encrypted_response2, IsOkAndHolds(EqualsProto(someExpectedEncryptedResponse2)));
```
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 33 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZgGPa8RdntthswPkj:-MZZgGPa8RdntthswPkk:b-f4qfod) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7e40ea70610d7943ed631d16ec5ccf023f4eacbe/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L33)):*
> ```cpp
> namespace wfa::panelmatch {
> namespace {
> using google::protobuf::RepeatedPtrField;
> ```

Please fully qualify (with leading "::")
___
*[src/test/kotlin/wfanet/panelmatch/protocol/common/CommutativeEncryptionUtilityTest.kt, line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZh8Mm9HA4SAKP4yPF:-MZZh8Mm9HA4SAKP4yPG:bim593n) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7e40ea70610d7943ed631d16ec5ccf023f4eacbe/src/test/kotlin/wfanet/panelmatch/protocol/common/CommutativeEncryptionUtilityTest.kt#L51)):*
> ```Kotlin
>         .build()
>     val encryptedResponse1: ApplyCommutativeEncryptionResponse =
>       JniCommutativeEncryption().applyCommutativeEncryption(encryptRequest1)
> ```

Maybe pull out some helper functions for the operations? Like:

```
fun applyCommutativeEncryption(key: ByteString, plaintexts: List<ByteString>): ByteString {
  val request = ApplyCommutativeEncryptionRequest.newBuilder().setEncryptionKey(key).addAllPlaintexts(plaintexts).build()
  return JniCommutativeEncryption().applyCommutativeEncryption(request)
}
```

This should make the test a little more pleasant to read without obscuring too much logic.

efoxepstein (commented):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 14 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: 18 of 55 files reviewed, 13 unresolved discussions (waiting on @efoxepstein and @yunyeng)

___
*[build/wfa/repositories.bzl, line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZb1W20hyrnSnhCLYK:-MZu-88R6mF3advRrQOe:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/build/wfa/repositories.bzl#L1)):*

For this and all the other files in `build`, are these all copied directly from `cross-media-measurement`? Are we able to just take a dependency on `cross-media-measurement` (for now) and refer to these files in that repo?
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZbfO4EqRbqqzDqNUu:-MZrvPFV59X6J68FR3MX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L30)):*

A using-declaration in a header file means that we're exporting the symbol as part of the public API of the library.

Since we don't expect anyone to write "wfanet::panelmatch::common::crypto::RepeatedPtrField", it's probably best to avoid the using-declaration and just spell out the full type name.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZbu_w-U71ukcCRNcY:-MZrvOLg91hK7cWr39Qu:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L54)):*

These can be removed, I think. There's no real point to private methods in an interface, is there? Nothing could call them.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZcH7c-mUwbW1xjr7Z:-MZrvNFG3qYsnEFY3frN:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L118)):*

Rather than duplicate this with above, let's just pull out the switch to a helper function.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.h, line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZcRCz4j9ym52zhFvI:-MZrqtAX-_COA89T5GL9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.h#L26)):*

I'm guessing this is copied from the other repository, but while you're here, we should fix it up.

The style guide prefers input parameters before outputs: "When ordering function parameters, put all input-only parameters before any output parameters." (https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs)

</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.cc, line 17 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZd38R8m9oaEv8ZBWO:-MZrqu718eGVEiiJrkU2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/encryption_utility_helper.cc#L17)):*

Delete this file?
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/started_thread_cpu_timer.h, line 26 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZd90i6WcQ4-WO1vct:-MZu-T7i4gve8elwvYnJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/common/crypto/started_thread_cpu_timer.h#L26)):*

Can we just take a dependency on some other repository instead of copying this stuff over? Should be better to maintain. Eventually, we'll pull out all the common stuff to a utility repo.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.h, line 23 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdMSu3qFvq-9sdjHX:-MZrqvS79TNG3B23ZPpK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.h#L23)):*

See previous comment on using-declarations in headers... this one could be considered OK, but since each of these things is only referenced once, doesn't really seem worthwhile to add the using-declarations.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdmpT8dQDKBUtIpvd:-MZrtTto5fgWtXeofplj:b-8w8p4u) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L51)):*

I would guess that you could write `ASSIGN_OR_RETURN(*response.mutable_whatever(), cryptor->Whatever())` -- does that work?
</blockquote></details>

I don't think that works or I'm misunderstanding. Do the built-in protobuf functions support an assignment like that?
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_wrapper.cc, line 45 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZeQAz3-HRhtwYui-N:-MZt7r_aDvayG9b-rFwU:bjmuzv0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_wrapper.cc#L45)):*

These feel like they're begging to be templated:

```
template <typename RequestT, typename ResponseT>
absl::StatusOr<std::string> SomeGoodName(const std::string& serialized_request, absl::FunctionRef<ResponseT(RequestT)> fn) {
  RequestT request_proto;
  RETURN_IF_ERROR(ParseRequestFromString(request_proto, serialized_request));
  ASSIGN_OR_RETURN(ResponseT response, fn(request_proto));
  return response.SerializeAsString(); 
}

absl::StatusOr<std::string> ReApplyCommutativeEncryptionWrapper(
    const std::string& serialized_request) {
  return SomeGoodName(serialized_request, ReApplyCommutativeEncryption);
}
```

I dunno, what do you think?
</blockquote></details>

I was having some trouble with this one and getting it to a place where it was succint and not more trouble than its worth. Let's just defer this one.
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZa_8Sen3uGotJs4W2-r1-96:-MZtLjlw594Oq2ilvHDD:b-3fkzfa) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/0bcdccd97a851ab5463a63c6c44456aa32ef257c/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L96)):*

https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/test/cc/testutil/matchers.h#L48

```
ASSERT_THAT(encrypted_response2, IsOkAndHolds(EqualsProto(someExpectedEncryptedResponse2)));
```

</blockquote></details>

I looked at the API and it looks like that doesn't work if the message types are different (which they are in this case). I did it a different way I'm happy with using an iterator but simpler than before.
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 33 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZgGPa8RdntthswPkj:-MZt9P4aAfasCUC4yJXb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7e40ea70610d7943ed631d16ec5ccf023f4eacbe/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L33)):*

Please fully qualify (with leading "::")
</blockquote></details>

Done.
___
*[src/test/kotlin/wfanet/panelmatch/protocol/common/CommutativeEncryptionUtilityTest.kt, line 51 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZh8Mm9HA4SAKP4yPF:-MZtM6R0-xcbWDDMCO7A:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7e40ea70610d7943ed631d16ec5ccf023f4eacbe/src/test/kotlin/wfanet/panelmatch/protocol/common/CommutativeEncryptionUtilityTest.kt#L51)):*

Maybe pull out some helper functions for the operations? Like:

```
fun applyCommutativeEncryption(key: ByteString, plaintexts: List<ByteString>): ByteString {
  val request = ApplyCommutativeEncryptionRequest.newBuilder().setEncryptionKey(key).addAllPlaintexts(plaintexts).build()
  return JniCommutativeEncryption().applyCommutativeEncryption(request)
}
```

This should make the test a little more pleasant to read without obscuring too much logic. 
</blockquote></details>

Done

efoxepstein (approved):
> Reviewed 37 of 37 files at r5, 5 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 6 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 45 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuRZn3B1rg5VyIYIRU:-MZuRZn3B1rg5VyIYIRV:bg6o5jb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L45)):*
> ```objc
>       Action action) = 0;
> 
>   virtual absl::StatusOr<::google::protobuf::RepeatedPtrField<std::string>>
> ```

Nit: no need to fully qualify here. https://abseil.io/tips/130 and https://abseil.io/tips/119 give some guidance here -- the TL;DR of what's consistent with our codebase is:

- using-declarations should always be fully qualified
- otherwise, try to avoid full qualification

(In other words, it's fine to just write "google::protobuf::RepeatedPtrField" just like you write "std::string" instead of "::std::string".)
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuSIL08f2tHjZdGWT5:-MZuSIL08f2tHjZdGWT6:bl1ugyg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L72)):*
> ```cpp
>       case Action::kEncrypt: {
>         return Encrypt(text);
>         break;
> ```

No need for break statements, they're unreachable. (And then since there's only one statement per case, you don't need the {}s either.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 85 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuSDYK5bBFeYaw92dt:-MZuSDYK5bBFeYaw92du:bjwu3s9) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L85)):*
> ```cpp
>         return absl::InvalidArgumentError("Unknown action.");
>     }
>     return result;
> ```

This line of code looks unreachable. I think you can just remove it (and the declaration for result).
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdmpT8dQDKBUtIpvd:-MZuSlgjEe2swRSz7U74:b8blov6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L51)):*

I don't think that works or I'm misunderstanding. Do the built-in protobuf functions support an assignment like that?
</blockquote></details>

`response.mutable_encrypted_texts()` is a RepeatedPtrField<std::string>*. (That is, a pointer to a RepeatedPtrField.)

So `*response.mutable_encrypted_texts()` is a RepeatedPtrField<std::string>&. (That is, a mutable reference to a RepeatedPtrField).

https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.repeated_field#RepeatedPtrField says that RepeatedPtrField has an assignment operator, so I would expect ASSIGN_OR_RETURN to work with it. Maybe there's some bug in the macro where the "*" causes problems? I'd have to see the specific error to debug more.

It's fine to leave as is, too.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 62 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuUFZs047jyIuQ8lm5:-MZuUFZs047jyIuQ8lm6:b5zkyz1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L62)):*
> ```cpp
> }
> 
> absl::StatusOr<::wfanet::panelmatch::protocol::protobuf::
> ```

(See other comment about full qualification.)
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZa_8Sen3uGotJs4W2-r1-96:-MZuWAfD5J87BUEwPV0H:bf7p0af) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/0bcdccd97a851ab5463a63c6c44456aa32ef257c/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L96)):*

I looked at the API and it looks like that doesn't work if the message types are different (which they are in this case). I did it a different way I'm happy with using an iterator but simpler than before.
</blockquote></details>

I feel like you probably want to make sure they're pointwise different -- otherwise maybe it just changes the first one and the rest are the same, for example. Try:

```
EXPECT_THAT(encrypted_texts2, Pointwise(Ne(), encrypted_texts1));
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h, line 45 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuRZn3B1rg5VyIYIRU:-MZwvOPC-lNZCanvrkrS:bhnize6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.h#L45)):*

Nit: no need to fully qualify here. https://abseil.io/tips/130 and https://abseil.io/tips/119 give some guidance here -- the TL;DR of what's consistent with our codebase is:

- using-declarations should always be fully qualified
- otherwise, try to avoid full qualification

(In other words, it's fine to just write "google::protobuf::RepeatedPtrField" just like you write "std::string" instead of "::std::string".)
</blockquote></details>

Makes sense. Not quite black and white but I'll try to adopt these styles ... feel free to keep giving nits
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 72 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuSIL08f2tHjZdGWT5:-MZwvzjK2VkxeGH-nWy1:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L72)):*

No need for break statements, they're unreachable. (And then since there's only one statement per case, you don't need the {}s either.
</blockquote></details>

Done.
___
*[src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc, line 85 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuSDYK5bBFeYaw92dt:-MZwvkMR5zmL0V_jLgPC:b-7z5f4h) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/common/crypto/cryptor.cc#L85)):*

This line of code looks unreachable. I think you can just remove it (and the declaration for result).
</blockquote></details>

yeah - noticed that after I submitted. Thanks for catching it too!
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 51 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZdmpT8dQDKBUtIpvd:-MZwyXlF7CmJm2HsEt-x:b-gsiapy) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/9901ba123ca6def1b1ff28f6ba7c7eaffd8d030a/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L51)):*
> SSIGN_OR_RETURN(*response.mutable_whatever(), cryptor->Whatever())
Thanks for explaining further. Works fine with ASSIGN_OR_RETURN as is.
___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 62 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuUFZs047jyIuQ8lm5:-MZwwsiw3Hqd7mRKlijf:b-9l9mbm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L62)):*

(See other comment about full qualification.)
</blockquote></details>

I agree on the other one but shouldn't this one stay the way it is? what if someone else puts together a wfanet namespace... Seems less exclusive than std or google for example.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 5 unresolved discussions (waiting on @efoxepstein and @yunyeng)

___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZa_8Sen3uGotJs4W2-r1-96:-MZx0Ixo7S5MLkWzSVGG:bn4u2a9) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/0bcdccd97a851ab5463a63c6c44456aa32ef257c/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L96)):*
> https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.repeated_field#RepeatedPtrField
how is that better than just ... they both check the range?
`return std::equal(arg.begin(), arg.end(), expected.begin(), expected.end());`

I went ahead and made this change but aren't they doing the same thing?

efoxepstein (approved):
> Reviewed 5 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 62 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuUFZs047jyIuQ8lm5:-MZxPpPI8eg_qmiXc3ah:b-ktlsio) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L62)):*

I agree on the other one but shouldn't this one stay the way it is? what if someone else puts together a wfanet namespace... Seems less exclusive than std or google for example.
</blockquote></details>

It shouldn't cause any problems unless it somehow got transitively included into this file and had the exact same symbol names that we refer to here. That's not gonna happen, I don't think. But they could still make a top level wfanet namespace and cause just as much chaos.

The convention we've been following is to avoid fully qualifying unless in a using-decl -- at a minimum, we should be consistent here and revisit as a broader topic if you disagree with the Abseil tips-of-the-week.
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 96 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZZa_8Sen3uGotJs4W2-r1-96:-MZxQeSa8jv0RkCbOYKI:beup3r6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/0bcdccd97a851ab5463a63c6c44456aa32ef257c/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L96)):*

> https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.repeated_field#RepeatedPtrField
how is that better than just ... they both check the range?
`return std::equal(arg.begin(), arg.end(), expected.begin(), expected.end());`

I went ahead and made this change but aren't they doing the same thing?
</blockquote></details>

1. Pointwise has good debug info. It will say exactly what elements and their positions differ.
2. [0, 1, 2, 3, 4, 5] and [9, 1, 2, 3, 4, 5] will pass the `Not HasEqualRepeatedFields` test. However, this probably shouldn't pass -- even though they are different, most items are the same. We probably care about a stronger property: that each element is different.
3. It's less surprising to the reader to see the standard, build-in matchers. They don't have to go searching for what HasEqualRepeatedFields actually does because they (should) be familiar with Pointwise.
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 109 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZxRM_f9Ihef_MU_Wh0:-MZxRM_f9Ihef_MU_Wh1:bmogsuj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/59de486f2db0c7c71c0472e0fea596253a3c47e9/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L109)):*
> ```cpp
>   ASSERT_THAT(decrypted_response1, IsOk());
>   auto decrypted_texts1 = (*decrypted_response1).decrypted_texts();
>   EXPECT_THAT(decrypted_texts1, Pointwise(Eq(), encrypted_texts2));
> ```

`Pointwise(Eq(), x)` is the same as the simpler `ElementsAre(x)`

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: all files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @yunyeng)

___
*[src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc, line 62 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZuUFZs047jyIuQ8lm5:-MZxSeVY1NTgMX82pxBN:b7a8ta3) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/bbe379dbad95d0ce3cf3f394c18005d131a7f6c5/src/main/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility.cc#L62)):*

It shouldn't cause any problems unless it somehow got transitively included into this file and had the exact same symbol names that we refer to here. That's not gonna happen, I don't think. But they could still make a top level wfanet namespace and cause just as much chaos.

The convention we've been following is to avoid fully qualifying unless in a using-decl -- at a minimum, we should be consistent here and revisit as a broader topic if you disagree with the Abseil tips-of-the-week.
</blockquote></details>

thats not exactly how I was read the Totw  but sounds good
___
*[src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc, line 109 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6#-MZxRM_f9Ihef_MU_Wh0:-MZxagHXAP7d-j-5_h9E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/59de486f2db0c7c71c0472e0fea596253a3c47e9/src/test/cc/wfanet/panelmatch/protocol/crypto/commutative_encryption_utility_test.cc#L109)):*

`Pointwise(Eq(), x)` is the same as the simpler `ElementsAre(x)`
</blockquote></details>

Done.

efoxepstein (approved):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yunyeng)

yunyeng (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #49, opened 2021-07-16
Add FHE batch lookup pipeline.

efoxepstein:
This is for issue #24.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)


SanjayVas (changes requested):
> Reviewed 1 of 7 files at r1, 5 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupQueryEvaluator.kt, line 80 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf039uw-Kd3vaM7JuGW:-Mf039uw-Kd3vaM7JuGX:b3zbtr1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupQueryEvaluator.kt#L80)):*
> ```
> Throws if
> ```
nit: use @throws KDoc tag indicating the exception type
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt, line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf05ifr-X2F_cth7iTW:-Mf05ifr-X2F_cth7iTX:b682sek) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt#L1)):*
> ```Kotlin
> // Copyright 2021 The Cross-Media Measurement Authors
> ```

While only public symbols need to have KDocs, I think that the private functions/types would benefit from having them as there's a decent amount of functional and generic type shenanigans.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt, line 110 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf05D_U6lu2uGXPHK0C:-Mf05D_U6lu2uGXPHK0D:b-5zo4kq) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt#L110)):*
> ```
> Pair<T.() -> V1, Coder<V1>>
> ```
I think having a specific class would help readability here rather than using Pair. If nothing else, it means that each usage would have one fewer type parameter specifications.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Pipeline.kt, line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf0A4MG0cpqkUdSMQPu:-Mf0A4MG0cpqkUdSMQPv:b-ali2p0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Pipeline.kt#L93)):*
> ```
> Exposed for testing.
> ```
Generally you only expose otherwise hidden implementation details if it helps testing some of the public API of the SUT. It's commonly done for things like constants and small utility functions.

In this case, it appears it's being exposed *so* it can be tested directly. This smells like testing an implementation detail rather than doing all testing via the public API.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt, line 1 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf05ifr-X2F_cth7iTW:-Mf3qoOL-h3rQDjBtxj6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt#L1)):*

While only public symbols need to have KDocs, I think that the private functions/types would benefit from having them as there's a decent amount of functional and generic type shenanigans.
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt, line 110 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf05D_U6lu2uGXPHK0C:-Mf3qowT7IDrZEGhiFF5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Coders.kt#L110)):*

> ```
> Pair<T.() -> V1, Coder<V1>>
> ```
I think having a specific class would help readability here rather than using Pair. If nothing else, it means that each usage would have one fewer type parameter specifications.
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Pipeline.kt, line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mf0A4MG0cpqkUdSMQPu:-Mf3wclc48xR6221Tizi:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/825a2f3ed9a4bccdd465cf961212ea73329c8260/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Pipeline.kt#L93)):*

> ```
> Exposed for testing.
> ```
Generally you only expose otherwise hidden implementation details if it helps testing some of the public API of the SUT. It's commonly done for things like constants and small utility functions.

In this case, it appears it's being exposed *so* it can be tested directly. This smells like testing an implementation detail rather than doing all testing via the public API.
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 1 of 7 files at r2, 12 of 15 files at r3, 1 of 1 files at r4, 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/BUILD.bazel, line 10 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfAOpD0DHLgHrddDjd9:-MfAOpD1EmXRQiF-v9MA:b-ue1i2b) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/70ee2ec88959811de026b3b91dca5466d5a51836/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/BUILD.bazel#L10)):*
> ```
>     testonly = True,
> ```
nit: You can just put default_testonly = True in the package declaration. Doesn't make a big difference in single-target packages like this, but a handy habit to get into for testing packages if they can end up with multiple targets.

stevenwarejones (changes requested):
> Reviewed 1 of 7 files at r1, 1 of 7 files at r2, 4 of 15 files at r3, 3 of 4 files at r5, 9 of 10 files at r6, 4 of 8 files at r7, 1 of 2 files at r8, 7 of 14 files at r9, 9 of 11 files at r10.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 12 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 40 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcsr2V8g-gLqhbv96b:-Mfcsr2V8g-gLqhbv96c:b-zghpf0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L40)):*
> ```Kotlin
> 
>   @Test
>   fun executeQueries() {
> ```

can you title this test more specifically?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 97 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfctHmv64Ei4xCWov3h:-MfctHmv64Ei4xCWov3i:b3hwor7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L97)):*
> ```Kotlin
> 
>   @Test
>   fun `combineResults single result`() {
> ```

add a `combineResults multiple result` test
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt, line 24 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcvcno2szKelqsDvZq:-Mfcvcno2szKelqsDvZr:b-7pqfk6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt#L24)):*
> ```Kotlin
> import org.wfanet.panelmatch.client.batchlookup.Result
> 
> object PlaintextQueryEvaluator : QueryEvaluator {
> ```

can you add some comments how this works?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt, line 40 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcu7U54r4Con-tR-xn:-Mfcu7U54r4Con-tR-xo:b-y7yh7h) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt#L40)):*
> ```Kotlin
>   }
> 
>   override fun combineResults(results: Sequence<Result>): Result {
> ```

in the real implementation, is this where the dot product of the results is done using FHE?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluatorTestHelper.kt, line 26 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcvhPl7L0GSxEJdeQ9:-MfcvhPl7L0GSxEJdeQA:bw153hw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluatorTestHelper.kt#L26)):*
> ```Kotlin
> import org.wfanet.panelmatch.client.batchlookup.ShardId
> 
> object PlaintextQueryEvaluatorTestHelper : QueryEvaluatorTestHelper {
> ```

add some comments how this works
___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 94 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTHvcA8PuZHZjdLQt:-MfhTHvcA8PuZHZjdLQu:b-fegem1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L94)):*
> ```protobuf
> 
> message BucketId {
>   int32 id = 1;
> ```

I'm thinking this should be int64. Can't bucket size be the number of items in database * 2
___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 98 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTy2m5a4-i2wmkI4G:-MfhTy2m5a4-i2wmkI4H:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L98)):*
> ```protobuf
> 
> message QueryId {
>   int32 id = 1;
> ```

ditto
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-MfcpYo59B1I3VA6_ThE:b-jftnka) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*
> ```Kotlin
>     // -1L is 2^64-1 when interpreted as unsigned.
>     // (2^64-1) % 7 == 1, ((2^64-1) / 7) % 3 == 2
>     assertThat(bucketing.apply(-1L)).isEqualTo(ShardId(1) to BucketId(2))
> ```

should this throw an error?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-MfhP4M99dq8htcbTHGK:bfl68jv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*
> ```Kotlin
>       }
>       .groupByKey("Group by Shard")
>       .parDo {
> ```

Whats the performance hit of doing it this way vs using `GroupIntoBatches`? Feels like an alternative could be to use GroupIntoBatches but set a conservative lower limit.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractBatchLookupWorkflowEndToEndTest.kt, line 86 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhSMWX9oKkQUDy466_:-MfhSMWX9oKkQUDy466a:b11o1am) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractBatchLookupWorkflowEndToEndTest.kt#L86)):*
> ```Kotlin
>     val results = workflow.batchLookup(databasePCollection, queryBundlesPCollection)
>     val localHelper = helper // For Beam's serialization
>     assertThat(results).satisfies {
> ```

can you comment this assert?
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfhQZq48zM8Y8HbogfH:b-ah1x9y) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*
> ```Kotlin
> 
> @RunWith(JUnit4::class)
> class BatchLookupWorkflowTest : BeamTestBase() {
> ```

I'm thinking there should be at test for 
1. grouping into batches with a strict size limit and checking the memory allocation of it
2. a test that throws an error if the subshardsizebytes is too small to hold anything
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-MfhROSZBcIQ5_NpJ3Bi:bn8iv0a) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*
> ```Kotlin
>     // With `parameters`, we expect database to have a single shard with 5 subshards.
>     // The buckets should be 53 to "abc", 58 to "def", 71 to "hij", 85 to "klm".
>     val parameters = Parameters(numShards = 1, numBucketsPerShard = 100, subshardSizeBytes = 5)
> ```

does the subshardSizeBytes make a difference here? It would be nice to have too identical tests where the subshardSizeBytes result in different results.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 12 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfhW2C32VQCCUXNzye1:b-n95kdz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

I'm thinking there should be at test for 
1. grouping into batches with a strict size limit and checking the memory allocation of it
2. a test that throws an error if the subshardsizebytes is too small to hold anything
</blockquote></details>

What do you think about just deleting this test and relying on the AbstractBatchLookupWorkflowTest to provide coverage?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 12 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfhXfUi7b5IXtmFb_Q_:bhe5x66) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

What do you think about just deleting this test and relying on the AbstractBatchLookupWorkflowTest to provide coverage?
</blockquote></details>

do you mean the existing `AbstractBatchLookupWorkflowEndToEndTest.kt`? I don't see `AbstractBatchLookupWorkflowTest.kt`

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 12 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 94 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTHvcA8PuZHZjdLQt:-MfhXbM05-_TvYjZ1DJr:b717nx5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L94)):*

I'm thinking this should be int64. Can't bucket size be the number of items in database * 2
</blockquote></details>

Bucket IDs are per shard. If each shard has more than 2^31 buckets, we'll have other major problems.
___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 98 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTy2m5a4-i2wmkI4G:-MfhYBxk-Fv2ErPsuVKG:beb01yb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L98)):*

ditto
</blockquote></details>

We're anticipating like 10,000 queries. It's computationally infeasible to handle more than 2^31
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-MfhYRuQ6KlC3bbOm1SD:b-m4gdfk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

should this throw an error?
</blockquote></details>

No -- the values come from unsigned 64 bit integers in C++ (see the Fingerprinter class in common-cpp). We just need to treat them as unsigned here.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfhYMa0AunLAuCtp6gq:b-ojgklw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

do you mean the existing `AbstractBatchLookupWorkflowEndToEndTest.kt`? I don't see `AbstractBatchLookupWorkflowTest.kt`
</blockquote></details>

Yep that's what I mean

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 11 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 94 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTHvcA8PuZHZjdLQt:-MfhdhcvC3UkbjkHA9Ur:b-x2h5dz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L94)):*

Bucket IDs are per shard. If each shard has more than 2^31 buckets, we'll have other major problems.
</blockquote></details>

hmmm.... what if someone puts in terrible parameters like a single shard. Do we want to throw warnings? crash? What should be the expected behavior?
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfhd4UADQyzbRCq_KXK:b-svylu8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

No -- the values come from unsigned 64 bit integers in C++ (see the Fingerprinter class in common-cpp). We just need to treat them as unsigned here.
</blockquote></details>

why not use ulong then?
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfheS0z2uj7jCo0nPb4:b-t1kplg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

Yep that's what I mean
</blockquote></details>

This test is a little less `abstract` than that test, which I like. Why were thinking of deleting it? I'd still like to see more tests around subshardsizebytes (too small, under the limit)

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 11 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-MfhuNY-DoUSIx98VnHC:b-e0rik8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

why not use ulong then?
</blockquote></details>

It's possible that someone will want to use this library from Java. Java has no unsigned integers as far as I know. I'm not sure how the conversion would work. Is there a real problem here?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 11 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto, line 94 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhTHvcA8PuZHZjdLQt:-Mfi3cnC3zm3iwL1EDUN:bnxjow9) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/proto/wfa/panelmatch/client/batchlookup/query_evaluator.proto#L94)):*

hmmm.... what if someone puts in terrible parameters like a single shard. Do we want to throw warnings? crash? What should be the expected behavior?
</blockquote></details>

`BatchLookupWorkflow.Parameters` takes in an int `numBucketsPerShard`. This means the type system enforces that an int32 here is sufficient. The constructor of `BatchLookupWorkflow.Parameters` will throw an exception for obviously invalid settings -- basically, negative integers.

Since parameters will be changed very rarely, they should definitely have a human in the loop and should not solely rely on automated validation.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 11 unresolved discussions (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 40 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcsr2V8g-gLqhbv96b:-MfiAAbc14FKvSz-Yiy9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L40)):*

can you title this test more specifically?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 97 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfctHmv64Ei4xCWov3h:-MfiABEJ9CPSrzjh9QX5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L97)):*

add a `combineResults multiple result` test
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt, line 24 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcvcno2szKelqsDvZq:-MfiCJxR0auU5ag1QnR3:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt#L24)):*

can you add some comments how this works?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt, line 40 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfcu7U54r4Con-tR-xn:-MfiCKoE6B3mJcEwSUji:b-7n5ueg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluator.kt#L40)):*

in the real implementation, is this where the dot product of the results is done using FHE?
</blockquote></details>

`combineResults` will just be some of the FHE addition of the dot products (each subshard is a dot product and then subshard results are added together in `combineResults`)
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluatorTestHelper.kt, line 26 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcvhPl7L0GSxEJdeQ9:-MfiCuKu5qS0JXtvfUGF:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/PlaintextQueryEvaluatorTestHelper.kt#L26)):*

add some comments how this works
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-Mfi5R8P967kjZg5bwEf:bljcwk6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*

Whats the performance hit of doing it this way vs using `GroupIntoBatches`? Feels like an alternative could be to use GroupIntoBatches but set a conservative lower limit.
</blockquote></details>

Yep -- that's reasonable. `groupByKey` does incur a fairly hefty increase in I/O cost. Would you prefer I use GroupIntoBatches?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractBatchLookupWorkflowEndToEndTest.kt, line 86 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhSMWX9oKkQUDy466_:-Mfi7LLp00kNFrszdqXa:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractBatchLookupWorkflowEndToEndTest.kt#L86)):*

can you comment this assert?
</blockquote></details>

Done.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-Mfi4DFwEg35CJN3hL_u:b6xn8pt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

This test is a little less `abstract` than that test, which I like. Why were thinking of deleting it? I'd still like to see more tests around subshardsizebytes (too small, under the limit)
</blockquote></details>

I was considering deleting it because it requires reasoning about internal details of the workflow, like which values will end up in which buckets. The other test doesn't care about this and just validates that the results are correct. https://abseil.io/tips/135 feels related (but that's C++-specific). See also https://testing.googleblog.com/2015/01/testing-on-toilet-prefer-testing-public.html.

`AbstractBatchLookupWorkflowEndToEndTest` runs tests for a few different `subshardSizeBytes` values and ensures they all yield the same results.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-MfiKTOp8n1zgM33TIwH:b-1doedo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

does the subshardSizeBytes make a difference here? It would be nice to have too identical tests where the subshardSizeBytes result in different results.
</blockquote></details>

This is one reason that I kind of prefer the other test -- it doesn't get too hung up on the implementation details. There's no way to actually observe in a unit test what subshardSizeBytes does. One reasonable way to test this would be to add some Metrics (https://beam.apache.org/documentation/programming-guide/#metrics) -- what do you think?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 25 of 32 files reviewed, 11 unresolved discussions (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfiReDR1vAboQJrPdBu:b-ii907w) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

I was considering deleting it because it requires reasoning about internal details of the workflow, like which values will end up in which buckets. The other test doesn't care about this and just validates that the results are correct. https://abseil.io/tips/135 feels related (but that's C++-specific). See also https://testing.googleblog.com/2015/01/testing-on-toilet-prefer-testing-public.html.

`AbstractBatchLookupWorkflowEndToEndTest` runs tests for a few different `subshardSizeBytes` values and ensures they all yield the same results.
</blockquote></details>

@sanjayvas also curious about your thoughts on this -- I'm less experienced on testing in the Java ecosystem

stevenwarejones (changes requested):
> Reviewed 1 of 11 files at r10, 1 of 6 files at r11.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 27 of 32 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 125 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfj-hN_1jXV1I7Banwr:-Mfj-hN_1jXV1I7Banws:bqq56e5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L125)):*
> ```Kotlin
>       .isEqualTo(DecodedResult(queryId, ByteString.EMPTY))
> 
>     assertFails { runCombineResults(nonEmptyResult, nonEmptyResult) }
> ```

assert the exception type and message please
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfiw9ST2XrVJrTj8JyH:b628xsw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

It's possible that someone will want to use this library from Java. Java has no unsigned integers as far as I know. I'm not sure how the conversion would work. Is there a real problem here?
</blockquote></details>

we could make them use guava which is already a dependency in the project. It supports `UnsignedLong`. In terms of the problem, I think implicit conversion of from signed values to unsigned is not a expectation that people integrating into the system would have and might cause unforeseen issues that are hard for them to debug.  

If we think that will be the case, then why not just reject negative numbers? 2^63 is a pretty big number.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-Mfj0BJkB6ZHQ4FXW8G7:b-uuztij) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*

Yep -- that's reasonable. `groupByKey` does incur a fairly hefty increase in I/O cost. Would you prefer I use GroupIntoBatches?
</blockquote></details>

I'm thinking that fitting that last little bit into each subshard probably isn't worth it. Two questions:
1. How much does it matter to maximize the size of each subshard (eg the benefit)
2. How often would GroupIntoMatches greatly exceed a maximum if we set the limit lower?
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-Mfizbw70GOrJTecbCJW:bwrcsko) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

This is one reason that I kind of prefer the other test -- it doesn't get too hung up on the implementation details. There's no way to actually observe in a unit test what subshardSizeBytes does. One reasonable way to test this would be to add some Metrics (https://beam.apache.org/documentation/programming-guide/#metrics) -- what do you think?
</blockquote></details>

so, we'd create a metric of size of each subshard?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 27 of 32 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 125 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfj-hN_1jXV1I7Banwr:-Mfj16EV7dBzpdUaNLq_:bpa8a6h) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L125)):*

assert the exception type and message please
</blockquote></details>

Actually, this test case is invalid in general -- since under FHE we can't tell if something is empty or not. I need to delete it.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-Mfj2zvgA7k1xE9cLOnC:bb778k5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

so, we'd create a metric of size of each subshard?
</blockquote></details>

We could create a `Distribution` that gives summary statistics of the sizes of subshards.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 27 of 32 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 125 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfj-hN_1jXV1I7Banwr:-Mfj4VXW6ujITaenyYN2:b-kq70gm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L125)):*

Actually, this test case is invalid in general -- since under FHE we can't tell if something is empty or not. I need to delete it.
</blockquote></details>

thats an interesting point. So, just refactor the test for unknownIfEmpty Results and combine them regardless
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-Mfj4gik930LKjWsb8VC:bg4ls84) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

We could create a `Distribution` that gives summary statistics of the sizes of subshards.
</blockquote></details>

How much work do you think that would be? Easy to add in or a follow-up?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 27 of 32 files reviewed, 5 unresolved discussions (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt, line 125 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-Mfj-hN_1jXV1I7Banwr:-MfjIfBl11K_8n8kRZ2V:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/testing/AbstractQueryEvaluatorTest.kt#L125)):*

thats an interesting point. So, just refactor the test for unknownIfEmpty Results and combine them regardless
</blockquote></details>

Done.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfj5DXf0fKb4FWuZKy1:bpstfc6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

we could make them use guava which is already a dependency in the project. It supports `UnsignedLong`. In terms of the problem, I think implicit conversion of from signed values to unsigned is not a expectation that people integrating into the system would have and might cause unforeseen issues that are hard for them to debug.  

If we think that will be the case, then why not just reject negative numbers? 2^63 is a pretty big number.
</blockquote></details>

I think we need to be consistent with https://github.com/world-federation-of-advertisers/any-sketch-java/blob/main/src/main/java/org/wfanet/anysketch/fingerprinters/Fingerprinter.java#L26 here. If you want to go through the effort of refactoring that to use UnsignedLong, go for it. I think we avoided using it there because of the overhead of using objects instead of primitives (and we wanted AnySketch to be fast).

Another thing to consider: the proto compiler for java turns both uint64 and int64 into Java's long. Since the crypto happens in C++, we'll be pulling the longs right out of protos.

How about I file an issue somewhere to come up with a consistent story for dealing with unsigned integers on the JVM but we leave this alone for now?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-MfjDRpI5Zg55lTNtB_l:bid6rzw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*

I'm thinking that fitting that last little bit into each subshard probably isn't worth it. Two questions:
1. How much does it matter to maximize the size of each subshard (eg the benefit)
2. How often would GroupIntoMatches greatly exceed a maximum if we set the limit lower?
</blockquote></details>

1. Not super important
2. No idea! Probably not too frequently.

Would it be alright to just add a TODO to switch to GroupIntoBatches in this PR? I'd like to get this one merged because it'll be a fair amount of work to switch it over.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-MfjDM2sAMmU4id9r8Ge:b4fztlx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

How much work do you think that would be? Easy to add in or a follow-up?
</blockquote></details>

I would prefer to do it in a follow-up PR.

SanjayVas (approved):
> Reviewed 9 of 10 files at r6, 1 of 8 files at r7, 4 of 14 files at r9, 8 of 11 files at r10, 6 of 6 files at r11, 2 of 2 files at r12.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 6 unresolved discussions (waiting on @efoxepstein, @oliver-amzn, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt, line 29 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfjD4l5Ah1rXsOHZ07f:-MfjD4l66l6mnwa363yW:b-mdiqwh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt#L29)):*
> ```
> databaseShardOf
> ```
I probably would have done with `buildShard` since this is just wrapping a builder, but this naming has its merits too.

(The "Of" suffix makes me think specifically of collection utils. Not sure if that's good or bad.)
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfjFgCr7qFJbTptgGH9:b-2ajusd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

@sanjayvas also curious about your thoughts on this -- I'm less experienced on testing in the Java ecosystem
</blockquote></details>

That Abseil tip applies equally well to Java, even if the specific ways around it don't. 

There's basically two concepts covered by those tips:
1. Test using the public API of the subject under test (SUT) and verify only the the documented/intended behavior.
2. Don't test implementation details. Sometimes you split out a different class/object/function for organizational purposes, but aren't necessarily creating a distinct SUT. The blog post linked above indicates tips on how to determine if something is just an implementation detail.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 6 unresolved discussions (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt, line 29 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfjD4l5Ah1rXsOHZ07f:-MfjK0o18sgeGmehRzxN:b-qfp4e7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt#L29)):*

> ```
> databaseShardOf
> ```
I probably would have done with `buildShard` since this is just wrapping a builder, but this naming has its merits too.

(The "Of" suffix makes me think specifically of collection utils. Not sure if that's good or bad.)
</blockquote></details>

How do you feel about functions that are the same names as types? E.g. https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex.html

I wasn't sure if `DatabaseShard(...)`, `databaseShardOf(...)`, or `buildDatabaseShard(...)` would be cleanest.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfjMnb-82Rp6hid1nFJ:bc9yguz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

That Abseil tip applies equally well to Java, even if the specific ways around it don't. 

There's basically two concepts covered by those tips:
1. Test using the public API of the subject under test (SUT) and verify only the the documented/intended behavior.
2. Don't test implementation details. Sometimes you split out a different class/object/function for organizational purposes, but aren't necessarily creating a distinct SUT. The blog post linked above indicates tips on how to determine if something is just an implementation detail.
</blockquote></details>

I guess I can't really argue that `BatchLookupWorkflowTest` should be removed -- it does test just the public API. But it is a little more brittle than `AbstractBatchLookupWorkflowEndToEndTest` as it requires understanding how sharding/bucketing happens. This isn't really an implementation detail because clients of `BatchLookupWorkflow` need to understand how bucketing works -- but it is very low level. I dunno -- I guess I'll leave the test if @stevenwarejones prefers.

oliver-amzn (approved):
> Reviewed 1 of 6 files at r11.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 31 of 32 files reviewed, 6 unresolved discussions (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

SanjayVas (approved):
> Reviewed 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 6 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt, line 29 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfjD4l5Ah1rXsOHZ07f:-MfjTJEJ6Pv0G_1RxfaD:bp2rnv1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt#L29)):*

How do you feel about functions that are the same names as types? E.g. https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.sync/-mutex.html

I wasn't sure if `DatabaseShard(...)`, `databaseShardOf(...)`, or `buildDatabaseShard(...)` would be cleanest.
</blockquote></details>

The `Mutex` factory function is an odd case as `Mutex` is actually an interface defined in an extensions library. The factory function is intentionally trying to look like a constructor to work around this fact. I wouldn't apply that naming scheme here.

Both of the other naming patterns have their merits. Both naming schemes make it decently obvious what the function does. `buildDatabaseShard` fits more with conventional function naming using a verb phrase, but there's precedent for names like `databaseShardOf`. The latter is more discoverable if your instinct is to look for a constructor before realizing it's a protobuf message.

Basically, either of the latter two naming schemes are fine with me. I was just leaving a non-blocking comment indicating that what you chose wouldn't have been my instinctive choice for the naming scheme.

oliver-amzn (approved):
> Reviewed 2 of 15 files at r3, 9 of 10 files at r6, 1 of 8 files at r7, 4 of 14 files at r9, 8 of 11 files at r10, 2 of 6 files at r11, 2 of 2 files at r12, 1 of 1 files at r13.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 6 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 5 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt, line 29 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfjD4l5Ah1rXsOHZ07f:-MfnqHv37AfCFaVAJrnF:b6r0e5x) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/aa828a9d537da0d541d1a3c9a8fdfa32638a872e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/Protos.kt#L29)):*

The `Mutex` factory function is an odd case as `Mutex` is actually an interface defined in an extensions library. The factory function is intentionally trying to look like a constructor to work around this fact. I wouldn't apply that naming scheme here.

Both of the other naming patterns have their merits. Both naming schemes make it decently obvious what the function does. `buildDatabaseShard` fits more with conventional function naming using a verb phrase, but there's precedent for names like `databaseShardOf`. The latter is more discoverable if your instinct is to look for a constructor before realizing it's a protobuf message.

Basically, either of the latter two naming schemes are fine with me. I was just leaving a non-blocking comment indicating that what you chose wouldn't have been my instinctive choice for the naming scheme.
</blockquote></details>

I see. I think I'm going to leave it as is for this PR, but I'll meditate on it and potentially rename in another PR.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-MfnuFPrDmdhzHEDGCZM:b-yrq3be) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

I think we need to be consistent with https://github.com/world-federation-of-advertisers/any-sketch-java/blob/main/src/main/java/org/wfanet/anysketch/fingerprinters/Fingerprinter.java#L26 here. If you want to go through the effort of refactoring that to use UnsignedLong, go for it. I think we avoided using it there because of the overhead of using objects instead of primitives (and we wanted AnySketch to be fast).

Another thing to consider: the proto compiler for java turns both uint64 and int64 into Java's long. Since the crypto happens in C++, we'll be pulling the longs right out of protos.

How about I file an issue somewhere to come up with a consistent story for dealing with unsigned integers on the JVM but we leave this alone for now?
</blockquote></details>

I'm good with an issue but also a comment here pointing to that url from any-sketch would be nice too
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-MfntmQZ4qJ0Mo9z2zrt:bwdq627) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*

1. Not super important
2. No idea! Probably not too frequently.

Would it be alright to just add a TODO to switch to GroupIntoBatches in this PR? I'd like to get this one merged because it'll be a fair amount of work to switch it over.
</blockquote></details>

okay - add as TODO
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-Mfnv06B4EQLF5H7H2gT:b-utkf2k) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

I guess I can't really argue that `BatchLookupWorkflowTest` should be removed -- it does test just the public API. But it is a little more brittle than `AbstractBatchLookupWorkflowEndToEndTest` as it requires understanding how sharding/bucketing happens. This isn't really an implementation detail because clients of `BatchLookupWorkflow` need to understand how bucketing works -- but it is very low level. I dunno -- I guess I'll leave the test if @stevenwarejones prefers.
</blockquote></details>

if we're okay with leaving this as-is, I'll resolve this one. I think my earlier comments are addressed with the separate metrics thread.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-MfnusKe5eTxougZBgaw:b-gora07) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

I would prefer to do it in a follow-up PR.
</blockquote></details>

file an issue please. I think any company implementing this reference implementation pipeline out-of-the-box will want these metrics.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 33 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhQZq48zM8Y8HbogfG:-MfnvzMe657_QdgPqGmq:bcrnyg4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L33)):*

if we're okay with leaving this as-is, I'll resolve this one. I think my earlier comments are addressed with the separate metrics thread.
</blockquote></details>

Yep, let's resolve this one

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt, line 109 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhROSZBcIQ5_NpJ3Bh:-MfnzqhM8bJBn4CGA9ME:b-87yo3l) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflowTest.kt#L109)):*

file an issue please. I think any company implementing this reference implementation pipeline out-of-the-box will want these metrics.
</blockquote></details>

https://github.com/world-federation-of-advertisers/panel-exchange-client/issues/66

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfo0Tfb2Z2mJKQ257hd:b-87yo3k) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

I'm good with an issue but also a comment here pointing to that url from any-sketch would be nice too
</blockquote></details>

https://github.com/world-federation-of-advertisers/panel-exchange-client/issues/67
___
*[src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt, line 126 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfhP4M87oWl_SQ1CyI5:-Mfo-ySM3-QahlDl78_F:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/091e864cb33dc5e28d02039186bf86a212be938e/src/main/kotlin/org/wfanet/panelmatch/client/batchlookup/BatchLookupWorkflow.kt#L126)):*

okay - add as TODO
</blockquote></details>

Done.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 1 unresolved discussion (waiting on @efoxepstein, @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfo1teTB5Gz3uBuWg_d:brduyik) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

https://github.com/world-federation-of-advertisers/panel-exchange-client/issues/67
</blockquote></details>

can you copy the body of what you wrote above into the issue? or post a link to it?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, 1 unresolved discussion (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt, line 57 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49#-MfcpYo59B1I3VA6_ThD:-Mfo3qME8pRL5kO0hW-l:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/47b8d7890167337b894b50877a0913f552e28322/src/test/kotlin/org/wfanet/panelmatch/client/batchlookup/BucketingTest.kt#L57)):*

can you copy the body of what you wrote above into the issue? or post a link to it?
</blockquote></details>

Done.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/49)* status: 30 of 32 files reviewed, all discussions resolved (waiting on @oliver-amzn, @SanjayVas, and @stevenwarejones)

PR #65, opened 2021-07-29
Change the Panel Storage abstraction to use Flows

jonmolle:
preparation for signed url storage, we don't want to arbitrarily collect
some of the data being passed as it will be too large to store in
memory. This allows us to collect the results as-needed.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)


efoxepstein (commented):
> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 4 unresolved discussions (waiting on @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnERV-D8m3lLKtro3p:-MfnERV-D8m3lLKtro3q:b-eeqep7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L44)):*
> ```
> require
> ```

Use `requireNotNull`?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnDREe1mzlK37895ih:-MfnDREe1mzlK37895ii:b6tl4qt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L45)):*
> ```Kotlin
> 
>     require(
>       inMemoryStorage.putIfAbsent(
> ```

Can you use `computeIfAbsent` to avoid processing the flow when unnecessary?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnEJNI2Tc_zM7uQV1c:-MfnEJNI2Tc_zM7uQV1d:b77hr6u) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L48)):*
> ```Kotlin
>         key,
>         try {
>           data.reduce { a, b -> a.concat(b) }
> ```

Can you use `fold` instead of `reduce` to avoid the try-catch? https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/fold.html
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/RemoteUrlStorageClient.kt, line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnEbCD2hgdcMDYHtxu:-MfnEbCD2hgdcMDYHtxv:bju5rcn) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/RemoteUrlStorageClient.kt#L63)):*
> ```Kotlin
>     check(operationType == "write") {
>       """
>         Not allowed to read from a URL tagged as in $operationType mode. Signed Urls only support
> ```

Should we have two interfaces? `Storage` and `MutableStorage` (like List and MutableList)? So we don't need to override unsupported methods?

jonmolle:
requireNotNull has an issue where if you write an empty ByteString (which the unit tests support so I assume is allowed) it will not be able to tell the difference between an existing key with empty value and a missing key. If you don't want to support null writes, then I think adding a check above the step is a safer option.

I think computeIfAbsent does not support the kind of operation we're trying to do here, but a DIY putIfAbsent seems to be fine for supporting empty ByteStrings and avoiding collecting the flow when we can't write it.

Finally, I didn't realize I had added that remote storage file, apologies. It does support both, just not both at once for the same url. I'm moving enforcement outside the class and will just gracefully handle if someone has plugged in workflows incorrectly.

efoxepstein (commented):
> Reviewed 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle)

efoxepstein (commented):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 44 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnERV-D8m3lLKtro3p:-MfoelSW8ejVTGHIg6Zt:bajzt9u) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L44)):*

> ```
> require
> ```

Use `requireNotNull`?
</blockquote></details>

refactored, this isn't really relevant anymore.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnDREe1mzlK37895ih:-MfodZuWESpuDia_kyFB:b5ol37z) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L45)):*

Can you use `computeIfAbsent` to avoid processing the flow when unnecessary?
</blockquote></details>

computeIfAbsent (and getOrPut, which was discussed offline) doesn't provide a way to know if we're overwriting an element or not (which is not allowed in the abstraction). I've changed it to a much simpler require(key is not in map) check followed by a simple assignment.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt, line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnEJNI2Tc_zM7uQV1c:-MfobjJ71xqRe_3B0Yjl:bb6un1j) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/InMemoryStorage.kt#L48)):*

Can you use `fold` instead of `reduce` to avoid the try-catch? https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/fold.html
</blockquote></details>

Done
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/RemoteUrlStorageClient.kt, line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-MfnEbCD2hgdcMDYHtxu:-Mfobw0R0sQr4cjyAoA_:b-cuvy1i) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/f3b01f3b963e4d9b98dbfa8652fc4b1d0ec89e72/src/main/kotlin/org/wfanet/panelmatch/client/storage/RemoteUrlStorageClient.kt#L63)):*

Should we have two interfaces? `Storage` and `MutableStorage` (like List and MutableList)? So we don't need to override unsupported methods?
</blockquote></details>

The interface should support both reads and writes, just not for the same url. As I said before, this wasn't really supposed to be in this commit and I've removed it into a followup PR for now. It turns out those catches weren't really necessary and can be handled in the input labels.

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg6ui_Y7V0jsKbj1bFb:b-edekro) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*
> ```Kotlin
>    */
>   @Throws(NotFoundException::class)
>   suspend fun batchRead(inputLabels: Map<String, String>): Map<String, ByteString> =
> ```

what do you think about having this, too, return flows... `Map<String, Flow<ByteString>>`
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6vxkZ7ABJ7LPcHVJK:-Mg6vxkZ7ABJ7LPcHVJL:b43lmot) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15832e23a346f3a00748f6bd985a9d84e90ca4c7/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*
> ```
>     withContext(Dispatchers.IO) {
>       coroutineScope {
> ```
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6wj6e3UflAaHqHsa4:-Mg6wj6e3UflAaHqHsa5:b-xxd1mw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*
> ```Kotlin
>             }
>           }
>           .mapValues { entry -> entry.value.await() }
> ```

is there an efficiency to reducing this before awaiting vs after awaiting?

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg6xCw2E8kmpQrwpLnZ:bgkqw4c) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*

what do you think about having this, too, return flows... `Map<String, Flow<ByteString>>`
</blockquote></details>

Even better, it would be good to use something like Blob (https://github.com/world-federation-of-advertisers/common-jvm/blob/main/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L39) so that it can be truly lazily read (in case it's not used at all).

But I feel like this can go in a follow-up PR. WDYT @stevenwarejones ?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6vxkZ7ABJ7LPcHVJK:-Mg6xh2WBgJAnAoENDsC:b5uyhnx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15832e23a346f3a00748f6bd985a9d84e90ca4c7/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

> ```
>     withContext(Dispatchers.IO) {
>       coroutineScope {
> ```

</blockquote></details>

What's that?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg6yyNG2ydSGTxP6rLv:b-cgpbx8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*

Even better, it would be good to use something like Blob (https://github.com/world-federation-of-advertisers/common-jvm/blob/main/src/main/kotlin/org/wfanet/measurement/storage/StorageClient.kt#L39) so that it can be truly lazily read (in case it's not used at all).

But I feel like this can go in a follow-up PR. WDYT @stevenwarejones ?
</blockquote></details>

I'm fine with follow-up or now. Can we add a TODO around this?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6vxkZ7ABJ7LPcHVJK:-Mg6y_wW3ldb76Eu9P6S:b-j2df5k) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15832e23a346f3a00748f6bd985a9d84e90ca4c7/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

What's that?
</blockquote></details>

copy paste user input on my side. ignore.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6wj6e3UflAaHqHsa4:-Mg6zCm_5JaLaoRYSQ4w:bbqc7m0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

is there an efficiency to reducing this before awaiting vs after awaiting?
</blockquote></details>

if you just return a Map of Flows, you can ignore this, but otherwise, I'd like to know your thinking.

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg6xe5m-oX0rLSYznxf:b-il8cqq) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*

I'm fine with follow-up or now. Can we add a TODO around this?
</blockquote></details>

It absolutely should return flows, otherwise the problem we're attempting to solve (oversize inputs) will just be moved here. I was intending to do this in a followup PR as Eli said, though. I was sort of assuming that this read would be used immediately, but that's a reasonable change. I don't see why you would use the batch commands otherwise.

That being said, if we're using Blob and backing everything with Storage Clients, then we can probably get away with not even using Storage.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6wj6e3UflAaHqHsa4:-Mg6yv1j-Fc7_mItfuDO:b-j5wiyz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

if you just return a Map of Flows, you can ignore this, but otherwise, I'd like to know your thinking.
</blockquote></details>

I think returning a Flow or similar system (IE Eli's Blob suggestion) is the way to go, since these returning the full inputs will almost certainly cause memory overflows which we're trying to avoid. I just hadn't really touched these because I was focused on getting my own new storage implementation working and I wasn't sure how these were intended to be used (I don't think any consumers have been fleshed out yet). 
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6vxkZ7ABJ7LPcHVJK:-Mg7-_CfBpCJo_0rrNfl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15832e23a346f3a00748f6bd985a9d84e90ca4c7/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

copy paste user input on my side. ignore.
</blockquote></details>

Done.

efoxepstein (approved):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @efoxepstein, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg78AnF0VUZprS4jzZe:b-w83hj1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*

It absolutely should return flows, otherwise the problem we're attempting to solve (oversize inputs) will just be moved here. I was intending to do this in a followup PR as Eli said, though. I was sort of assuming that this read would be used immediately, but that's a reasonable change. I don't see why you would use the batch commands otherwise.

That being said, if we're using Blob and backing everything with Storage Clients, then we can probably get away with not even using Storage.
</blockquote></details>

okay - can you either add a TODO or migrate to Blob?

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 54 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6ui_Y7V0jsKbj1bFa:-Mg7XKTG7sAEqc_M20L4:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L54)):*

okay - can you either add a TODO or migrate to Blob?
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 5 of 7 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

efoxepstein (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

jonmolle (commented):
> Dismissed @stevenwarejones from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 63 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/65#-Mg6wj6e3UflAaHqHsa4:-Mg8ESDT1hnIKEmgWGVU:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/089a7d71907cfa469be02b78934ea1c3a00c37a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L63)):*

I think returning a Flow or similar system (IE Eli's Blob suggestion) is the way to go, since these returning the full inputs will almost certainly cause memory overflows which we're trying to avoid. I just hadn't really touched these because I was focused on getting my own new storage implementation working and I wasn't sure how these were intended to be used (I don't think any consumers have been fleshed out yet). 
</blockquote></details>

Done.

PR #89, opened 2021-08-16
Deprecate and remove the Storage class

### LINKED ISSUE #81: Is it worth replacing Storage with extension methods on top of StorageClient?
After some implementation, there isn't really much that Storage does past batchRead and batchWrite. Other methods are passthrough to the underlying StorageClient.

It may be possible to just do all of what Storage offers in a few extension methods. This is already done in the case of Blob.toByteString().

Something like 
StorageClient.batchRead / batchWrite

jonmolle:
Resolves #81

- This moves to using the underlying StorageClient classes directly.
- functions like batchRead/Write have been made extensions, with the intent
to migrate them into StorageClient as-needed.
- Added a stub function for each verifiedRead/Write, with a downstream
PR to implement the new signing/verification code (as well as a PR to
move some of that code's dependencies into common- repos.)
- Deletes: InMemoryStorage, Storage (class)
- FileSystemStorage remains due to the init code not having a place to
migrate yet. Once that has a home in either a constructor for
StorageClient or init function for exchange steps.

unit tests refactored to cover the new code.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)


efoxepstein (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 31 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFYY_P4HbF2-3ddrxH:-MhFYY_P4HbF2-3ddrxI:b-367lwm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L31)):*
> ```Kotlin
> import org.wfanet.measurement.storage.read
> 
> class StorageNotFoundException(inputKey: String) : Exception("$inputKey not found")
> ```

OOC, why not continue to use `null` to signal that something isn't found? Is there a reason we prefer throwing for verified operations?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWdrs9G4IDsJdyhcf:-MhFWdrs9G4IDsJdyhcg:b-rvyjzv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L47)):*
> ```Kotlin
>   //  validation process using a provided [X509Certificate]. Right now it just runs [getBlob].
>   //  This function assumes that StorageClient will handle tracking the [X509Certificate].
>   println(blobKey)
> ```

Remove `println`?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWqw14Yvy6Ol-TeqY:-MhFWqw14Yvy6Ol-TeqZ:b-cjseot) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L75)):*
> ```Kotlin
>   return withContext(Dispatchers.IO) {
>     coroutineScope {
>       inputLabels
> ```

There's actually potential for a data race here because `inputLabels` could actually change during the execution of this coroutine. Guava's ImmutableMap would protect against this, as would deep copying the map or acquiring a lock on the Map.

Not sure what, if anything, is best -- just thought I'd flag this.

It also happens below, too.

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 31 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFYY_P4HbF2-3ddrxH:-MhFZiq4-t8egY_bMKZY:bt6mhpf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L31)):*

OOC, why not continue to use `null` to signal that something isn't found? Is there a reason we prefer throwing for verified operations?
</blockquote></details>

Yes, specifically the InputTask wants to catch a specific error to know that something isn't ready. There might be a better way to do that, but it seems reasonable to me and this ports the functionality
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 47 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWdrs9G4IDsJdyhcf:-MhFZxnbD-EJi--Wrjo-:boonzr5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L47)):*

Remove `println`?
</blockquote></details>

whoops. Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWqw14Yvy6Ol-TeqY:-MhF_bPe7qnBeMSy8xyu:b-aos6px) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L75)):*

There's actually potential for a data race here because `inputLabels` could actually change during the execution of this coroutine. Guava's ImmutableMap would protect against this, as would deep copying the map or acquiring a lock on the Map.

Not sure what, if anything, is best -- just thought I'd flag this.

It also happens below, too.
</blockquote></details>

I didn't realize., I thought the builtin kotlin maps were immutable by default and you had to use MutableMap otherwise. Same with the result of mapValues(). Do you have a preference? I'll work on this right now.

efoxepstein (changes requested):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 1 unresolved discussion (waiting on @jonmolle, @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWqw14Yvy6Ol-TeqY:-MhFcLya--QHCHXPbanA:b-7084v) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L75)):*

I didn't realize., I thought the builtin kotlin maps were immutable by default and you had to use MutableMap otherwise. Same with the result of mapValues(). Do you have a preference? I'll work on this right now.
</blockquote></details>

`Map` is immutable but `MutableMap` is a `Map` so you can pass a `MutableMap` to a function that accepts a `Map`, then mutate it (outside of that function).

stevenwarejones (changes requested):
> Reviewed 3 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @efoxepstein, @jonmolle, @kgaynor, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 31 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKg51N5TaiMlkQKjnU:-MhKg51N5TaiMlkQKjnV:b-ycph7y) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L31)):*
> ```Kotlin
> import org.wfanet.measurement.storage.read
> 
> class StorageNotFoundException(inputKey: String) : Exception("$inputKey not found")
> ```

does this belong in the consent-signaling repo or measurement in general? Seems like a use case that others are dealing with, such as signed sketches.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhKggNuANDpjGGfZXqU:bsdapay) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*
> ```Kotlin
>  */
> @Throws(StorageNotFoundException::class)
> suspend fun StorageClient.verifiedRead(blobKey: String): Blob {
> ```

Needs to pass in the certificate as input to this function.

Also, how are you planning on getting the signature out of blob storage? Is the blob going to be read as a Flow<SignedData> or will the downstream function check the SignedData and output bytes?

Either way, the downstream needs a certificate to verify against.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 55 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKgu237BN3GHsEdE4A:-MhKgu237BN3GHsEdE4B:b-2fxr60) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L55)):*
> ```Kotlin
>  * other party using a pre-provided [X509Certificate].
>  */
> suspend fun StorageClient.verifiedWrite(blobKey: String, content: Flow<ByteString>): Blob {
> ```

should pass in a privateKeyHandle and a certificate
https://github.com/world-federation-of-advertisers/consent-signaling-client/blob/main/src/main/kotlin/org/wfanet/measurement/consent/crypto/Utils.kt#L27

efoxepstein (approved):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhLLJO29Fz8_XNk9z6G:b4w63wt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*

Needs to pass in the certificate as input to this function.

Also, how are you planning on getting the signature out of blob storage? Is the blob going to be read as a Flow<SignedData> or will the downstream function check the SignedData and output bytes?

Either way, the downstream needs a certificate to verify against.
</blockquote></details>

I'm wondering if we should have:

```
class SignedStorage(private val verifyCertificate, private val signCertificate, private val storage: Storage): Storage {
  override fun read(...) {
    // read payload and signature from `storage` and verify with `verifyCertificate`
  }

  override fun write(...) {
    // write to `storage` while computing signature with `signCertificate` and then write the signature
  }
}
```

stevenwarejones (changes requested):
> Reviewed 1 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @jonmolle, @kgaynor, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhLTDwk2uHYKx2KQjow:bc7ld3g) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*

I'm wondering if we should have:

```
class SignedStorage(private val verifyCertificate, private val signCertificate, private val storage: Storage): Storage {
  override fun read(...) {
    // read payload and signature from `storage` and verify with `verifyCertificate`
  }

  override fun write(...) {
    // write to `storage` while computing signature with `signCertificate` and then write the signature
  }
}
```
</blockquote></details>

Ideally, it should just pass in a reference to get it from kvm. 

So, the SignedStorage class would represent a two way relationship? Eg party1 reads and writes to a particular storage assumed to only be read and written by party2? I don't think the thing I like about it is its hard-codes in a two-party exchange.

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhMa7aM7UG2bkhyVQIE:b-npjtj5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*

Ideally, it should just pass in a reference to get it from kvm. 

So, the SignedStorage class would represent a two way relationship? Eg party1 reads and writes to a particular storage assumed to only be read and written by party2? I don't think the thing I like about it is its hard-codes in a two-party exchange.
</blockquote></details>

Hmm, not sure -- maybe we'd just want `SignedStorageReader` and SignedStorageWriter` abstractions. Anyway, that's a topic for another PR. This PR is deleting unnecessary code--I think we just need a TODO somewhere to move stuff into the consent repo.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @jonmolle, @kgaynor, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhOPKAnAN0nwun5DArT:b-i25j1j) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*

Hmm, not sure -- maybe we'd just want `SignedStorageReader` and SignedStorageWriter` abstractions. Anyway, that's a topic for another PR. This PR is deleting unnecessary code--I think we just need a TODO somewhere to move stuff into the consent repo.
</blockquote></details>

I'd be okay with separate `SignedStorageReader` and `Writer` abstractions. In some use cases, you might not want to construct a full read/write reader with private/public keys. Best to keep them separate.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 55 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKgu237BN3GHsEdE4A:-MhOQ9qG9oxl2YybkzQk:b-82243c) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L55)):*

should pass in a privateKeyHandle and a certificate
https://github.com/world-federation-of-advertisers/consent-signaling-client/blob/main/src/main/kotlin/org/wfanet/measurement/consent/crypto/Utils.kt#L27

</blockquote></details>

Can you add comments that a privateKeyHandle is needed, etc? and that maybe the entire VerifiedStorageClient should move to consent modeling?

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @kgaynor, and @yunyeng)

efoxepstein (approved):
> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @kgaynor, and @yunyeng)

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @kgaynor, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhFWqw14Yvy6Ol-TeqY:-MhPhBWd7gAbW1HMjoBj:b-3xm23b) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/624a08e0e3e676b56c93e93b081f2298fdbf4f05/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L75)):*

`Map` is immutable but `MutableMap` is a `Map` so you can pass a `MutableMap` to a function that accepts a `Map`, then mutate it (outside of that function).
</blockquote></details>

Done. I've enforced it as immutable and added a secondary check to avoid attempting to write twice to the same location. There are probably some checks down into StorageClient that could be done, but that can be another PR (and I think they should be minimal, if anything).
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 31 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKg51N5TaiMlkQKjnU:-MhPhXmy6WdzV2iSEsNE:b94jn64) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L31)):*

does this belong in the consent-signaling repo or measurement in general? Seems like a use case that others are dealing with, such as signed sketches.
</blockquote></details>

My plan is to temporarily add the new PrivateKey.sign(certificate: X509Certificate, data: Flow<ByteString>) and X509Certificate.verifySignature(data: Flow<ByteString>, signature: Flow<ByteString>) in panel exchange and consent signaling at the same time and then do a followup PR to use the consent signaling code (and remove the panel exchange version) once the consent signaling PR is in.

These function stubs exist to read (or write) both the data and signatures based on a single label value, they don't do any actual signing or verification themselves.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 42 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKggNuANDpjGGfZXqT:-MhPlm7o7dac1L-0ecd4:bxg1rfm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L42)):*

I'd be okay with separate `SignedStorageReader` and `Writer` abstractions. In some use cases, you might not want to construct a full read/write reader with private/public keys. Best to keep them separate.
</blockquote></details>

I'm happy with creating an abstraction in a downstream PR - there should probably be a place to keep the appropriate certificate(s).

Considering private and public storage are already separate in panel exchange, and the workflow itself is pretty strongly linked to a two-party exchange, I don't really see an issue with having a single abstraction that covers both cases, but I don't really have a strong opinion here. In theory for each step you should know which cert will be used to verify any specific input or output. It is possible to parameterize that, but in the short to medium term I think that's just giving ourselves more work for a possible future use-case. I also don't think changing things to add that functionality in later will be significantly more difficult than adding it in now.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 55 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89#-MhKgu237BN3GHsEdE4A:-MhPnddA-cs4OCKNXI3W:b-ccjff) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b9bf9db5bbdde8588f0fb02f1a5e3eb15cd6bb4f/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L55)):*

Can you add comments that a privateKeyHandle is needed, etc? and that maybe the entire VerifiedStorageClient should move to consent modeling?
</blockquote></details>

These are stubs, I'm going to add that in in the next PR.

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @kgaynor, @stevenwarejones, and @yunyeng)

stevenwarejones (approved):
> Reviewed 6 of 12 files at r1, 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/89)* status: all files reviewed, 3 unresolved discussions (waiting on @kgaynor, @stevenwarejones, and @yunyeng)

PR #98, opened 2021-08-26
add plaintext batch final decryption into existing batch decrypt

stevenwarejones:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)


stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 17 unresolved discussions (waiting on @efoxepstein)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2n1_D5-5b_unBPMHp:-MiMbemG80pKjvm4BjKB:be0e463) (no related file):*

I feel like 75% of this will disappear once we do everything in one JNI call. Is it worth just shooting directly for that instead of building out all this Kotlin?
</blockquote></details>

I thought since its already built-out here, there isn't much more to build out after this imo. I went ahead and built it to be able to fake it well. I think that and the tests help break down the different sub-functions, hopefully, more clearly to others who might look at this project in the future.

---
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/DecryptQueryResultsWorkflow.kt, line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2h4Dm9Mo6KKY3aq74:-MiMc5RaE0Laz9bZXbwX:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/DecryptQueryResultsWorkflow.kt#L57)):*

inline rhs and remove unnecessary variable?
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/Protos.kt, line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2howQ0Tkth45k34SU:-MiMeR5z051CagblfFYG:b-11honz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/Protos.kt#L52)):*

Please use strongly typed parameters (use QueryId and ShardId instead of Ints). (Test helpers can use ints but we really want to take advantage of typecheckers as much as possible in production)
</blockquote></details>

done. some of these were already merged up. I tried moving them all out to a protohelper file to distinguish which ones are okay in tests vs not.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/Protos.kt, line 70 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2hcAEBRLkrwpejNYL:-MiMe_6V09Y6ZjIqteES:bgjs2px) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/Protos.kt#L70)):*

I think the String versions of all of these should be in a testonly library. We really don't want to use them in production.
</blockquote></details>

agree. done 
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/SymmetricPrivateMembershipCryptor.kt, line 21 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2iG6W2v373Zsm5CT2:-MiMem9Q6RqPgsniBpV9:b-yktnyl) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/SymmetricPrivateMembershipCryptor.kt#L21)):*

I don't understand what you mean by symmetric here. The "Private Membership" problem has the security guarantees described here (as opposed to Private Information Retrieval).
</blockquote></details>

Maybe I'm just overstating it. The current plaintext private membership cryptor does not have that guarantee nor does the JNI one. One way to guarantee the data is computationally symmetric is to apply a second step of encrypting the data and making sure your querier only has keys to the data they have rights to. That is what I'm trying to explain here. There are other auditing techniques as well like the query padding mitigation but that is not detailed nor implemented yet.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/AbstractDecryptQueryResultsWorkflowTest.kt, line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2isvNB-dENTG21JHO:-MiMfTkb6Yyclim80ufj:bh6w2bl) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/AbstractDecryptQueryResultsWorkflowTest.kt#L101)):*

Is this the same as asserting that it is equal to `PLAINTEXTS`?
</blockquote></details>

correct. switched it to that.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextPrivateMembershipCryptorHelper.kt, line 80 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2lTzN0TxIIRSDvMSL:-MiMg3J_1n2ccAV_1Y0a:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextPrivateMembershipCryptorHelper.kt#L80)):*

Pair can be destructured, which might aid in readability:

```
  stuff.map { (plaintext, joinkey) -> ... }
```
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt, line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2lsvCEvKBAUysy-4W:-MiMgKJcCK6BCfu-dNl0:bkstkko) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt#L32)):*

Is there any situation in which we would want to override the privateMembershipCryptor or symmetricCryptor? (It seems dangerous)

If not, maybe just pull them out of the constructor to normal private instance variables?
</blockquote></details>

I put it in the constructor for two reasons:
1. Dependency injection
2. Flexibility if different symmetric cryptors were used in the future for different exchange

In what ways do you think that is dangerous? Do you think  someone might use an unapproved symmetric cryptor?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt, line 45 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2luue6zmdzXWbPmpk:-MiMgpWJ0PipEuJanl9f:bd30ifx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt#L45)):*

I think this is a no-op. Do you mean `publicKey = request.publicKey`?
</blockquote></details>

good catch
___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/SymmetricCryptor.kt, line 20 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2m_rH3qQ1jqjzQ_Z5:-MiMgwu73VeHijS5-pX2:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/SymmetricCryptor.kt#L20)):*

"`private`"
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/AbstractSymmetricCryptorTest.kt, line 28 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2mtev4e3oELwNkjRd:-MiMh1Uf2iSQKhtix8Vb:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/AbstractSymmetricCryptorTest.kt#L28)):*

`protected abstract val` probably
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt, line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jKtE4uAVUQH2aFei:-MiMh3_v3dE5vfHMWnf2:b5mxec7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt#L22)):*

Any reason we don't just concatenate them in tests?
</blockquote></details>

meaning a concat cryptor vs an xor cryptor?
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jlqR88xsh-vA_g55:-MiMhBhf2XKfsTyjizRv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L33)):*

Remove `ed`?
</blockquote></details>

Done.
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jrRq2p3CuiRrnBsf:-MiMhETa8O-2f4g0uoXH:bh4sf6w) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L40)):*

Remind me why we're doing this in two JNI calls instead of one?
</blockquote></details>

This is actually the proto for a single step. I left in the proto for doing it in 2 steps, too, but that wouldn't be part of the reference implementation.

I will follow up this PR very shortly with the c++ implementing this proto.
___
*[src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc, line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2kQmp4z-wKZCPOjim:-MiMkIJ28kXLCYvo207k:b7ki3wr) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc#L64)):*

It would be more typical to just write `test_request.mutable_single_blinded_joinkey()->set_key(key);` here
</blockquote></details>

done. should I do something similar for this line?
`test_request.mutable_encrypted_event_data()->CopyFrom(encrypted_data_batch);`
___
*[src/test/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptorTest.kt, line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2kuNx1TNe02qN9DqG:-MiMoEd64VxEbi4_H-cE:bughntg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/test/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptorTest.kt#L63)):*

`this.` not needed
</blockquote></details>

I'm pretty sure it is needed. That variable name was declared outside the proto.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptorTest.kt, line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2l01uDtKTnQMFIKA_:-MiMoBgvBc6S6qCwmMSB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/test/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptorTest.kt#L74)):*

Just assert that it is PLAINTEXTS?
</blockquote></details>

Done.

efoxepstein (changes requested):
> Reviewed 12 of 13 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: 29 of 30 files reviewed, 8 unresolved discussions (waiting on @duliomatos1, @efoxepstein, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt, line 32 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2lsvCEvKBAUysy-4W:-MiNP19c-kgFkeGc4YcJ:b-uqje3i) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/PlaintextSymmetricPrivateMembershipCryptor.kt#L32)):*

I put it in the constructor for two reasons:
1. Dependency injection
2. Flexibility if different symmetric cryptors were used in the future for different exchange

In what ways do you think that is dangerous? Do you think  someone might use an unapproved symmetric cryptor?
</blockquote></details>

This is fine -- it seems a little YAGNI to me but I can live with it
___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/ProtosHelper.kt, line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNPug90D3DmIVLsFb_:-MiNPug90D3DmIVLsFba:b-x9xo1i) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/739bc337e558f11c3dc041b3766a7a04fcdeefa0/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/ProtosHelper.kt#L15)):*
> ```Kotlin
> // limitations under the License.
> 
> package org.wfanet.panelmatch.client.privatemembership
> ```

Shouldn't the package match the path for consistency with other testing libraries? E.g. https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/main/src/main/kotlin/org/wfanet/panelmatch/common/beam/testing/Coders.kt#L15
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNQM3hCPun5y6iu6Kf:-MiNQM3i-rbR2f59PA01:b-ospu2k) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L39)):*
> ```protobuf
> 
> // Removes in one step both the private membership encryption as well as the
> // symmetric encryption that ensures the data is computationally symmetric.
> ```

I definitely don't understand what's meant by "computationally symmetric" -- can you elaborate?
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jrRq2p3CuiRrnBsf:-MiNQD2Y7lXpOhKACmXr:b-vavvdk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L40)):*

This is actually the proto for a single step. I left in the proto for doing it in 2 steps, too, but that wouldn't be part of the reference implementation.

I will follow up this PR very shortly with the c++ implementing this proto.
</blockquote></details>

I don't quite follow -- can we delete the unused proto? Or is it not unused?
___
*[src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc, line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2kQmp4z-wKZCPOjim:-MiNQSYu4twe3D9kp7_D:b9y1bfs) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc#L64)):*

done. should I do something similar for this line?
`test_request.mutable_encrypted_event_data()->CopyFrom(encrypted_data_batch);`
</blockquote></details>

Yes please

efoxepstein (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: 29 of 30 files reviewed, 6 unresolved discussions (waiting on @duliomatos1, @efoxepstein, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt, line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jKtE4uAVUQH2aFei:-MiNfogaDi6qVqJwMS4-:b7u03hh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt#L22)):*

meaning a concat cryptor vs an xor cryptor?
</blockquote></details>

Correct

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: 29 of 30 files reviewed, 5 unresolved discussions (waiting on @duliomatos1 and @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/ProtosHelper.kt, line 15 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNPug90D3DmIVLsFb_:-MiOODt09iklQpfAqfx9:b4h8s32) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/739bc337e558f11c3dc041b3766a7a04fcdeefa0/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/testing/ProtosHelper.kt#L15)):*

Shouldn't the package match the path for consistency with other testing libraries? E.g. https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/main/src/main/kotlin/org/wfanet/panelmatch/common/beam/testing/Coders.kt#L15
</blockquote></details>

yes. fixed.
___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt, line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jKtE4uAVUQH2aFei:-MiOOK_6DuPhXYEHpIDB:b-yr1j7t) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt#L22)):*

Correct
</blockquote></details>

I mean I don't think it really makes a difference. I can implement a concat cryptor instead if you prefer or leave as is.
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNQM3hCPun5y6iu6Kf:-MiOOdjg74YiDHgF-aFh:b-91boy8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L39)):*

I definitely don't understand what's meant by "computationally symmetric" -- can you elaborate?
</blockquote></details>

So, I've been going on this wiki page: https://en.wikipedia.org/wiki/Private_information_retrieval

As I understand it, Private Membership doesn't guarantee that the operation is Computationally Symmetric...eg guarantee the querier can't obtain information they shouldn't query. That's why we add in the layer of AES and the DBE to make it fully computationally symmetric. Happy to rename or provide more notes or a link to the wiki if you think that helps.
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jrRq2p3CuiRrnBsf:-MiOOzkaC7K4RuslQcXK:b-nkeo98) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L40)):*

I don't quite follow -- can we delete the unused proto? Or is it not unused?
</blockquote></details>

It will be used in the next PR. This is the proto for the single JNI call. You'll see it implemented in the next PR.
___
*[src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc, line 64 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2kQmp4z-wKZCPOjim:-MiOPzFK3moNtCbR9OZP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/test/cc/wfa/panelmatch/client/privatemembership/decrypt_event_data_test.cc#L64)):*

Yes please
</blockquote></details>

Done.

efoxepstein (approved):
> Reviewed 1 of 13 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1 and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt, line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jKtE4uAVUQH2aFei:-MiOW8K38gBklVpGrlql:b-sql8d8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt#L22)):*

I mean I don't think it really makes a difference. I can implement a concat cryptor instead if you prefer or leave as is.
</blockquote></details>

I'd prefer concat because it makes it easier to see what's going on in error messages (compare "salkfjas890fj108jsfoaisuj" with "<some data> encrypted with key <some key>")
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNQM3hCPun5y6iu6Kf:-MiOX2UG9C_YXc_pM30n:b-pvoqm4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L39)):*

So, I've been going on this wiki page: https://en.wikipedia.org/wiki/Private_information_retrieval

As I understand it, Private Membership doesn't guarantee that the operation is Computationally Symmetric...eg guarantee the querier can't obtain information they shouldn't query. That's why we add in the layer of AES and the DBE to make it fully computationally symmetric. Happy to rename or provide more notes or a link to the wiki if you think that helps.
</blockquote></details>

Ah, so it's not quite Computationally Symmetric PIR because it's also sparse. And Computationally Symmetric Sparse Private Information Retrieval is too awkward of a term and doesn't give much intuition on what it is. Instead, we use the term "private membership" or "private set membership" to mean the same thing.

Somewhere we should document that. I'd prefer consistency with github.com/google/private-membership but I'm also open to calling it Computationally Symmetric Sparse Private Information Retrieval if you want to document somewhere that it's the same thing as Private Membership.

(Why "set membership" instead of "map lookup" or something? It's straightforward to show that they're equivalent in functionality: if you had a system that only returned boolean responses, you could just query on the bits of the full response -- in other words, your query would be "does item x exist with bit i = 1?" But thankfully we don't need to do it that way in practice. And the membership term gives more intuition [to me at least] about the sparsity. There's also probably some historical reasons for the name -- it appears in the literature next to "private set intersection" a lot, a fairly common protocol.)

See also: https://github.com/google/private-membership/blob/main/README.md

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1 and @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt, line 22 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-Mi2jKtE4uAVUQH2aFei:-MiRrrFE1w9648I2CJFk:besjekj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/XorSymmetricCryptor.kt#L22)):*

I'd prefer concat because it makes it easier to see what's going on in error messages (compare "salkfjas890fj108jsfoaisuj" with "<some data> encrypted with key <some key>")
</blockquote></details>

Thought about it some more. Agree. Done
___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNQM3hCPun5y6iu6Kf:-MiRrx5f5sTkuIvCbdzW:biz1ggd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L39)):*

Ah, so it's not quite Computationally Symmetric PIR because it's also sparse. And Computationally Symmetric Sparse Private Information Retrieval is too awkward of a term and doesn't give much intuition on what it is. Instead, we use the term "private membership" or "private set membership" to mean the same thing.

Somewhere we should document that. I'd prefer consistency with github.com/google/private-membership but I'm also open to calling it Computationally Symmetric Sparse Private Information Retrieval if you want to document somewhere that it's the same thing as Private Membership.

(Why "set membership" instead of "map lookup" or something? It's straightforward to show that they're equivalent in functionality: if you had a system that only returned boolean responses, you could just query on the bits of the full response -- in other words, your query would be "does item x exist with bit i = 1?" But thankfully we don't need to do it that way in practice. And the membership term gives more intuition [to me at least] about the sparsity. There's also probably some historical reasons for the name -- it appears in the literature next to "private set intersection" a lot, a fairly common protocol.)

See also: https://github.com/google/private-membership/blob/main/README.md
</blockquote></details>

I kind of think our implementation is a little more nuanced than the private-membership readme. We aren't just returning if it's in the set or not and we're also hashing the values which allows collisions. Those factors do not make it, by itself, computationally symmetric. Without the AES encryption, the querier *can* learn about queried data outside of just learning if something is a member or not and also it can learn data its not supposed to be able to query.

However, at the core, what I'm really trying to distinguish is the feature set difference between just the vanilla PrivateMembershipCryptor that uses private-membership and what I'm currently calling the SymmetricPrivateMembershipCryptor that uses private-membership + aes to alleviate the problems with just using set membership how we're using it.

I'm not really sure how to name it. I'm okay naming it CSSPIR in the filename and elaborating in the readme.

efoxepstein (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 2 unresolved discussions (waiting on @duliomatos1 and @stevenwarejones)

___
*[src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiNQM3hCPun5y6iu6Kf:-MiS0eqyEfNJghZtE7Tm:b-pucwrd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/a55b146aac7cdc2e64677135aa6e187433221f53/src/main/proto/wfa/panelmatch/client/privatemembership/decrypt_event_data.proto#L39)):*

I kind of think our implementation is a little more nuanced than the private-membership readme. We aren't just returning if it's in the set or not and we're also hashing the values which allows collisions. Those factors do not make it, by itself, computationally symmetric. Without the AES encryption, the querier *can* learn about queried data outside of just learning if something is a member or not and also it can learn data its not supposed to be able to query.

However, at the core, what I'm really trying to distinguish is the feature set difference between just the vanilla PrivateMembershipCryptor that uses private-membership and what I'm currently calling the SymmetricPrivateMembershipCryptor that uses private-membership + aes to alleviate the problems with just using set membership how we're using it.

I'm not really sure how to name it. I'm okay naming it CSSPIR in the filename and elaborating in the readme. 
</blockquote></details>

Let's discuss this over Slack a bit. Can we merge this PR and rename in a follow-up?
___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiS0B4X1q8AduOWKksz:-MiS0B4X1q8AduOWKkt-:b-1dbq2d) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/c49bb2d3c59a1a55faba6491bd8dfb51667f8ede/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt#L25)):*
> ```Kotlin
> 
>   override fun encrypt(privateKey: ByteString, data: ByteString): ByteString {
>     return privateKey.concat(data)
> ```

The issue with this is `encrypt("1", "23")` and `encrypt("12", "3")` are identical. This is why I prefer `"<${data.toStringUtf8()}> encrypted with <${key.toStringUtf8()}>"`

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1 and @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt, line 25 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiS0B4X1q8AduOWKksz:-MiTJcWz-yy10_Aot5Zh:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/c49bb2d3c59a1a55faba6491bd8dfb51667f8ede/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt#L25)):*

The issue with this is `encrypt("1", "23")` and `encrypt("12", "3")` are identical. This is why I prefer `"<${data.toStringUtf8()}> encrypted with <${key.toStringUtf8()}>"`
</blockquote></details>

Done.

efoxepstein (changes requested):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1 and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt, line 33 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiTNSZH3fCTYf9AP4xW:-MiTNSZH3fCTYf9AP4xX:btjcy7o) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7d16d69f7c014c25b0689cbb2e6fe908b15dcb38/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt#L33)):*
> ```Kotlin
>     return data
>       .toStringUtf8()
>       .substring(
> ```

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/remove-suffix.html is probably more elegant

```
val suffix = SEPARATOR + privateKey.toStringUtf8()
val dataString = data.toStringUtf8()
require(dataString.endsWith(suffix))
return dataString.removeSuffix(suffix).toByteString()
```

efoxepstein (approved):
We're close enough here, :lgtm:, merge when you're happy with it

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-:-MiTWdcqDFhooW-eKL5u:b-i0vqd1)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1 and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1 and @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt, line 33 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/98#-MiTNSZH3fCTYf9AP4xW:-MiTmTxC1R03f4zU-06h:bdhi0t8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/7d16d69f7c014c25b0689cbb2e6fe908b15dcb38/src/main/kotlin/org/wfanet/panelmatch/common/crypto/testing/ConcatSymmetricCryptor.kt#L33)):*

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/remove-suffix.html is probably more elegant

```
val suffix = SEPARATOR + privateKey.toStringUtf8()
val dataString = data.toStringUtf8()
require(dataString.endsWith(suffix))
return dataString.removeSuffix(suffix).toByteString()
```
</blockquote></details>

Done. Didn't know `removeSuffix` was a thing. thx!

PR #100, opened 2021-08-31
Adds a VerifiedStorageClient wrapper for StorageClient

jonmolle:
This is to ensure that all reads and writes generate, and validate
against, signatures. VerifiedStorageClient (and the VerifiedBlob member
class) are storage agnostics, expecting an initialized StorageClient as
an input. It adds tracking for 2 certificates (one for reading, one for
writing) and a private key used in the validation.

- This will depend on extension functions in consent-signaling-client
- This is also sort of a rollback for my refactoring out Storage.
- Adding the actual consent-signaling dependencies is in a downstream PR
to avoid being blocked on commit. As a result
VerifiedBlob.verifiedRead() and VerifiedStorageClient.createBlobI() are
still stubs.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)


efoxepstein (changes requested):
> Reviewed 14 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: 14 of 23 files reviewed, 2 unresolved discussions (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MiXWxgpEkkswmqsSbas:b-h9r5l2) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*
> ```Kotlin
>    */
>   suspend fun execute(input: Map<String, VerifiedBlob>): Map<String, Flow<ByteString>>
> }
> ```

Is it important that ExchangeTask knows the blob is verified or can we keep it as Blob here and handle verification transparently to this class?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXXiHX7ex-rP6TT8Do:-MiXXiHZ2r0fm_fZfYW9:b-vjqufr) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*
> ```Kotlin
>   }
> 
>   open class VerifiedBlob(
> ```

Can this be a Blob? Or are we going for some type safety benefit?

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: 14 of 23 files reviewed, 2 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MiX_OMe0LRjL6wCNZ4N:b-5x5uv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*

Is it important that ExchangeTask knows the blob is verified or can we keep it as Blob here and handle verification transparently to this class?
</blockquote></details>

The issue is that to verify the blob you have to read the flow, which involves reading in from whatever underlying storage system there is (which we don't support, you have to get the blob again, which may not work with every storage implementation).

By passing the blob out without verification you're giving the caller the option to use the verification extension or not, which both gives more options for ignoring that safety valve and is surprisingly easy to do (since the verifiedRead extension isn't even in the same repo as StorageClient).

verifiedBlob is just a wrapper around Blob that invisibly verifies that the data matches the appropriate signature. I can rename it, since technically the blob isn't verified until it is actually read (which it is not at the point this line of code is run), but I don't recommend changing this back to just Blob.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXXiHX7ex-rP6TT8Do:-MiXcZqZCf-3IfSMCcdr:b-b316bb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*

Can this be a Blob? Or are we going for some type safety benefit?
</blockquote></details>

Hmm, this could be a Blob subclass, but I don't really want to expose the underlying StorageClient (from the underlying Blob) nor do I really want to make VerifiedStorageClient a StorageClient subclass. Mostly because I want to make it extremely clear that we expect all this stuff to be signed and to not let someone sneak around that step (whether in good faith or not) without significant effort.

It seems like a really easy thing to cause breakages, as VerifiedStorageClient creates (and expects) functionally invisible signature files to exist and if one party changes how that is handled it can cause a bunch of errors.

If you don't think that will be an issue I don't really mind changing that, they are effectively subclasses of StorageClient and Blob

efoxepstein (changes requested):
> Reviewed 9 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 9 unresolved discussions (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MiYJ-Yi9COSbNLHr7j8:bfvmlmp) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*

The issue is that to verify the blob you have to read the flow, which involves reading in from whatever underlying storage system there is (which we don't support, you have to get the blob again, which may not work with every storage implementation).

By passing the blob out without verification you're giving the caller the option to use the verification extension or not, which both gives more options for ignoring that safety valve and is surprisingly easy to do (since the verifiedRead extension isn't even in the same repo as StorageClient).

verifiedBlob is just a wrapper around Blob that invisibly verifies that the data matches the appropriate signature. I can rename it, since technically the blob isn't verified until it is actually read (which it is not at the point this line of code is run), but I don't recommend changing this back to just Blob.
</blockquote></details>

I guess the alternative I was suggesting is:

```
class VerifyingBlob(private val blob: Blob) : Blob {
  override fun read(...) {
    return unverifiedBlob.read().verifyFlow(...)
  }
 ...
}
```

This would NOT give the caller an opportunity to read the underlying unverified blob without verifying it, right?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 36 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiY9eO365zZZl-4ZXej:-MiY9eO365zZZl-4ZXek:bgf76uk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L36)):*
> ```Kotlin
> class StorageNotFoundException(inputKey: String) : Exception("$inputKey not found")
> 
> open class VerifiedStorageClient(
> ```

Why `open`?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYI3EVD-GyBknnEx-W:-MiYI3EVD-GyBknnEx-X:b-kglox) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L64)):*
> ```Kotlin
>           .mapValues { entry ->
>             async(start = CoroutineStart.DEFAULT) {
>               this@VerifiedStorageClient.getBlob(blobKey = entry.value)
> ```

Do you mind refactoring this to just make it process the reads (and writes for batchWrite) in serial? I feel like we don't have a good enough reason to support async writes and there's risk of bugs if getBlob isn't threadsafe. So maybe just `return inputLabels.mapValues { getBlob(blobKey = entry.value) }`
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 111 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYITcI6gJJhUsQLIpk:-MiYITcI6gJJhUsQLIpl:b-i2hx0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L111)):*
> ```Kotlin
>    */
>   @Throws(StorageNotFoundException::class)
>   suspend fun getBlob(blobKey: String): VerifiedBlob {
> ```

Can this be private?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 125 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIVVFAtoKTumakV7o:-MiYIVVFAtoKTumakV7p:b-i2hx0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L125)):*
> ```Kotlin
>    * verified by the other party using a pre-provided [X509Certificate].
>    */
>   suspend fun createBlob(blobKey: String, content: Flow<ByteString>): VerifiedBlob {
> ```

Can this be private?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIXtCB7y6aPvIjOfQ:-MiYIXtCB7y6aPvIjOfR:b-b5t0tw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*
> ```Kotlin
>   }
> 
>   open class VerifiedBlob(
> ```

why `open`?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 146 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIZvLCWPadOivhflR:-MiYIZvLCWPadOivhflS:bttutk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L146)):*
> ```Kotlin
>   ) {
> 
>     fun size(): Long {
> ```

```
val size: Long
  get() = sourceBlob.size
```
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 150 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIcuv4SPxEKBSwlzd:-MiYIcuv4SPxEKBSwlze:b-89a4l7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L150)):*
> ```Kotlin
>     }
> 
>     fun defaultBufferSizeBytes(): Int {
> ```

Ditto
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 166 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIe-Y0K9AWc-Y9tg_:-MiYIe-Y0K9AWc-Y9tga:b-yrkvoe) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L166)):*
> ```Kotlin
>     }
> 
>     fun read(bufferSize: Int): Flow<ByteString> {
> ```

can you default the argument rather than providing an overload?

stevenwarejones (changes requested):
> Reviewed 11 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 11 unresolved discussions (waiting on @jonmolle and @yunyeng)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-MiZpcIp-X8sogBYo9s3:bvzpn2l) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*
> ```Kotlin
>       readCertificate(FIXED_SERVER_CERT_PEM_FILE),
>       readCertificate(FIXED_SERVER_CERT_PEM_FILE),
>       readPrivateKey(FIXED_SERVER_KEY_FILE, KEY_ALGORITHM)
> ```

for shared storage, you should generate separate certificates for the different parties that you are verifying... eg you need to pass in your own certificate to write to the storage but need the other parties certificate to do a verified read.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt, line 62 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZq6OV1yHblTr14l7P:-MiZq6OV1yHblTr14l7Q:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt#L62)):*
> ```Kotlin
>       underlyingSharedStorage,
>       readCertificate(FIXED_SERVER_CERT_PEM_FILE),
>       readCertificate(FIXED_SERVER_CERT_PEM_FILE),
> ```

ditto

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 11 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MiYoayz28SpZbuPsxjf:bfxtegb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*

I guess the alternative I was suggesting is:

```
class VerifyingBlob(private val blob: Blob) : Blob {
  override fun read(...) {
    return unverifiedBlob.read().verifyFlow(...)
  }
 ...
}
```

This would NOT give the caller an opportunity to read the underlying unverified blob without verifying it, right?
</blockquote></details>

That seems fine, but do you think giving the callers the option to feed it a different type of Blob from another source is okay? I guess if we enforce the usage of a verifying storage client that's fine.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 36 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiY9eO365zZZl-4ZXej:-MiYp6Tf8O6FqqU4dGUW:b-boqjpj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L36)):*

Why `open`?
</blockquote></details>

Same as the other question - I needed them able to be subclassed for how I was doing the unit tests, but changed the tests and forgot to change this back. Removed.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 64 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYI3EVD-GyBknnEx-W:-MiYpPsyAg7_cSOYqv0v:b-7cg3il) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L64)):*

Do you mind refactoring this to just make it process the reads (and writes for batchWrite) in serial? I feel like we don't have a good enough reason to support async writes and there's risk of bugs if getBlob isn't threadsafe. So maybe just `return inputLabels.mapValues { getBlob(blobKey = entry.value) }`
</blockquote></details>

No problem, done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 111 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYITcI6gJJhUsQLIpk:-MiYpabD11fQZFdmsfhG:bqldoi) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L111)):*

Can this be private?
</blockquote></details>

I guess so, but if someone wants to get a specific blob rather than using the batch functions this could be useful. If we want to force everyone through batch then absolutely. I just thought it might be useful to get a single specific blob.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 125 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIVVFAtoKTumakV7o:-MiYpjo-5I2g3Kq6wl8E:bfv0a2s) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L125)):*

Can this be private?
</blockquote></details>

Same answer as the getBlob()
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIXtCB7y6aPvIjOfQ:-MiYivyj4CJ4myuft6zD:b-6k3jkb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*

why `open`?
</blockquote></details>

Whoops. I was mocking it in an earlier version of the unit tests, but that turned out not to be necessary. I've changed that. 
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 146 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIZvLCWPadOivhflR:-MiYnu9G0XULh9QjYCjB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L146)):*

```
val size: Long
  get() = sourceBlob.size
```
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 150 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIcuv4SPxEKBSwlzd:-MiYoBSz1Eb4aNM2UZN6:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L150)):*

Ditto
</blockquote></details>

Done.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 166 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIe-Y0K9AWc-Y9tg_:-MiYoUv6DH5L1fC5Fwfx:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L166)):*

can you default the argument rather than providing an overload?
</blockquote></details>

Done.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mibpmu95H6HGOgCsvQL:b-yiwm2l) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

for shared storage, you should generate separate certificates for the different parties that you are verifying... eg you need to pass in your own certificate to write to the storage but need the other parties certificate to do a verified read.
</blockquote></details>

I agree, this is specifically for testing. The two certs being passed here represent the read certificate (the other party's, for verification) and write certificate (your own, for signing). They should be able to be the same one in the case of private storage (where you are the reader and the writer). I could use a different cert for the two, but since I'm going to be signing and reading the stuff in downstream I think having them be the same is ok, especially as these tests aren't intended to test the cryptography. I may add one test in the storage tests when those functions fill out to ensure the proper cert is used on reads and writes, but here it is unnecessary.
___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt, line 62 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZq6OV1yHblTr14l7P:-MibrnaX8gYIo-TVKxmR:b8ht5aa) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt#L62)):*

ditto
</blockquote></details>

Responded above.

efoxepstein (changes requested):
> Reviewed 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 5 unresolved discussions (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MicGgjr6lxZkUtBpQOj:b-r6ffss) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*

That seems fine, but do you think giving the callers the option to feed it a different type of Blob from another source is okay? I guess if we enforce the usage of a verifying storage client that's fine.
</blockquote></details>

Your call here.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIXtCB7y6aPvIjOfQ:-MicGCeP7cjm_lYsng_v:b-8zplzx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*

Whoops. I was mocking it in an earlier version of the unit tests, but that turned out not to be necessary. I've changed that. 
</blockquote></details>

Reviewable is still showing it as open
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 137 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MicGNwE9ftZPX1K933A:-MicGNwE9ftZPX1K933B:bcp74u5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/89b9b0d8ebf381da6b64c8e5eb8c63d300344764/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L137)):*
> ```Kotlin
>      */
>     private fun verifiedRead(bufferSize: Int): Flow<ByteString> {
>       return sourceBlob.read(bufferSize)
> ```

Do you mind adding a TODO here so it's super super clear this isn't done yet?

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: 22 of 23 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiXWxgm3MMZaMF14MQ5:-MicfFmY7LChrKfxZVEY:b6e8y5w) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L30)):*

Your call here.
</blockquote></details>

I have a mild preference for an enforced wrapper, so I'll keep it as-is. Partly to avoid potential multiple-inheritance issues between the storage client subclasses' own blob subclasses.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 140 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiYIXtCB7y6aPvIjOfQ:-MicffnNBOsd0LhyuGGq:b-88oy0q) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L140)):*

Reviewable is still showing it as open
</blockquote></details>

Sorry, it got lost on a local branch, it is in now.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 137 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MicGNwE9ftZPX1K933A:-MicfDUQ1SZzqq-HYgfQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/89b9b0d8ebf381da6b64c8e5eb8c63d300344764/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L137)):*

Do you mind adding a TODO here so it's super super clear this isn't done yet?
</blockquote></details>

Done.

efoxepstein (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

efoxepstein (changes requested):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 3 unresolved discussions (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MigB6tt-F5qvXzjeK5N:-MigB6tt-F5qvXzjeK5O:b-bj96hx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/74a8df0795a4225a971ec3a44df1bc5f7695f08e/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L81)):*
> ```Kotlin
>    * Notes:
>    * - The validation happens in a separate thread and is non-blocking, but will throw a terminal
>    * ```
> ```

This looks wrong?

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MigB6tt-F5qvXzjeK5N:-MigoIQV6u18HtbDtNJ4:b-45s6uu) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/74a8df0795a4225a971ec3a44df1bc5f7695f08e/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L81)):*

This looks wrong?
</blockquote></details>

I know, I tried removing it, but the linter complains if it isn't like this. I assume some markdown weirdness? I have been having trouble getting ktfmt running locally in the past few days, so I'm less than excited to muck about with a kdoc comment.

efoxepstein (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @jonmolle, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MigB6tt-F5qvXzjeK5N:-Migwh9N024pfqjd5g5i:b45418p) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/74a8df0795a4225a971ec3a44df1bc5f7695f08e/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L81)):*

I know, I tried removing it, but the linter complains if it isn't like this. I assume some markdown weirdness? I have been having trouble getting ktfmt running locally in the past few days, so I'm less than excited to muck about with a kdoc comment.
</blockquote></details>

You probably just had the wrong amount of indentation on the second line. Get ktfmt working and just put it all on one line then let the autoformatter fix.

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: 22 of 23 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt, line 81 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MigB6tt-F5qvXzjeK5N:-Mih4Xiu9xxZ6RYVoL6E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/74a8df0795a4225a971ec3a44df1bc5f7695f08e/src/main/kotlin/org/wfanet/panelmatch/client/storage/Storage.kt#L81)):*

You probably just had the wrong amount of indentation on the second line. Get ktfmt working and just put it all on one line then let the autoformatter fix.
</blockquote></details>

Done.

efoxepstein (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones and @yunyeng)

yunyeng (approved):
> Reviewed 20 of 23 files at r1, 2 of 3 files at r3, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

efoxepstein (approved):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 1 of 23 files at r1, 2 of 3 files at r3, 1 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mj0kPvw7eGRxZheR4t5:b5fy6cv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

I agree, this is specifically for testing. The two certs being passed here represent the read certificate (the other party's, for verification) and write certificate (your own, for signing). They should be able to be the same one in the case of private storage (where you are the reader and the writer). I could use a different cert for the two, but since I'm going to be signing and reading the stuff in downstream I think having them be the same is ok, especially as these tests aren't intended to test the cryptography. I may add one test in the storage tests when those functions fill out to ensure the proper cert is used on reads and writes, but here it is unnecessary.
</blockquote></details>

I'm thinking that maybe we need an enum type shared that requires different read and write certificates to prevent errors like this.

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mj0oyk_Epl9d_awnNWh:b-fu9no1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

I'm thinking that maybe we need an enum type shared that requires different read and write certificates to prevent errors like this.
</blockquote></details>

I'm not against it, but I'm curious what safety you suggest that formalizing the difference like that will provide past passing two certs in labeled as read & write.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mj1G6x62PQ87T1BS5uK:b-pylhyh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

I'm not against it, but I'm curious what safety you suggest that formalizing the difference like that will provide past passing two certs in labeled as read & write.
</blockquote></details>

Some of the other checks I'm thinking about like making sure the x509 cert matches up with the private key are probably best dealt with in underlying libraries.

So, thinking about this a little more, I think the interface feels a little confusing since we are using the writeCert just for the algorithm name but using the the readCert for both a public key and algorithm name.

However, one reason to check here is the unique interface. If someone misconfigured their storage incorrectly, it just might be confusing to the parties why party 1 is able to write signed data but when its checked against their public key, it doesn't verify correctly because of algorithm name mismatches.

What do you think?

jonmolle (commented):
> Reviewed 1 of 23 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mj1OzNxE_BkrcgXPYEH:bng3cfi) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

Some of the other checks I'm thinking about like making sure the x509 cert matches up with the private key are probably best dealt with in underlying libraries.

So, thinking about this a little more, I think the interface feels a little confusing since we are using the writeCert just for the algorithm name but using the the readCert for both a public key and algorithm name.

However, one reason to check here is the unique interface. If someone misconfigured their storage incorrectly, it just might be confusing to the parties why party 1 is able to write signed data but when its checked against their public key, it doesn't verify correctly because of algorithm name mismatches.

What do you think?
</blockquote></details>

I kind of get where you're coming from, but I prefer to both keep these as close to the existing signature functions (see my other PR) as possible, which means using the cert just for choosing the algorithm. On top of that, since you should be using the cert you've passed to the other party, you should be confident that any error means the certificate has been tampered with (either by you, accidentally, or a bad actor). Passing the algorithm around is just one more thing to remember (that you don't have to).

I don't mind changing the name, I have it that way for simplicity (writeCert is the cert used in the writing of things and should match your privateKey.) Perhaps I could change the language to refer to ownership rather than action?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt, line 47 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZpcIp-X8sogBYo9s2:-Mj2MS6z7j1vjAkLxsv-:b-8botxh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchangeTest.kt#L47)):*

I kind of get where you're coming from, but I prefer to both keep these as close to the existing signature functions (see my other PR) as possible, which means using the cert just for choosing the algorithm. On top of that, since you should be using the cert you've passed to the other party, you should be confident that any error means the certificate has been tampered with (either by you, accidentally, or a bad actor). Passing the algorithm around is just one more thing to remember (that you don't have to).

I don't mind changing the name, I have it that way for simplicity (writeCert is the cert used in the writing of things and should match your privateKey.) Perhaps I could change the language to refer to ownership rather than action?
</blockquote></details>

Lets revisit this later. lgtm

efoxepstein (approved):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

___
*[src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt, line 62 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/100#-MiZq6OV1yHblTr14l7P:-Mj5RI06CHdzzzOXYpYH:b-2k9gw4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/10f39cf2e2c6b5275b0ee672cc4e2139e32652a8/src/test/kotlin/org/wfanet/panelmatch/client/exchangetasks/InputTaskTest.kt#L62)):*

Responded above.
</blockquote></details>

Marking as resolved because above comment says we can revisit later.

PR #135, opened 2021-09-24
Introduce and use SecretSet abstraction for validating ExchangeWorkflows

efoxepstein:
It's a cleaner abstraction to just store a list of all serialized ExchangeWorkflows in a secret manager somewhere. This way we don't need to worry about key rotation.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/135)


stevenwarejones (approved):
> Reviewed 2 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/135)* status: all files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @jonmolle, and @yunyeng)

PR #138, opened 2021-09-27
Add remaining ExchangeTasks for Apache Beam workflows

efoxepstein:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138)


efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138)* status: 4 of 11 files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt, line 52 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138#-MkeEH6Q1AAPickD8JAd:-MkeFsC-AKnpxZKjSaLU:butaumy) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/5edd9d24afa047e9faa27e3b9c4ec7001f85ee4f/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt#L52)):*

don't you want to join these in a pipeline?
</blockquote></details>

Each blob contains a single SharedInputs proto, right? That means it needs to be loaded into memory anyway to be parsed. So why not join in memory, which will be orders of magnitude faster (provided it fits into ram).
___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/DecryptPrivateMembershipResultsTask.kt, line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138#-MkeDOMpALYtmZWo9vzN:-MkeGAjZEMjElH8WIKHm:b-inz66d) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/5edd9d24afa047e9faa27e3b9c4ec7001f85ee4f/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/DecryptPrivateMembershipResultsTask.kt#L74)):*

doesn't this come in from the agreed upon workflow?
</blockquote></details>

Sure, we could add it as a parameter to `ExchangeWorkflow.Step.DecryptPrivateMembershipQueryResultsStep` if you want. Can we let this TODO capture the Action Item to do that?
___
*[src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteSignedFiles.kt, line 58 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138#-MkeD5PRCSuz43-LdTB8:-MkeGSkQ0vODRxxFxVwB:b6radhc) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/5edd9d24afa047e9faa27e3b9c4ec7001f85ee4f/src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteSignedFiles.kt#L58)):*

can we make this more platform agnostic?
</blockquote></details>

This is a URI, not a path, so `/` should be the unambiguously correct separator, no?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/138)* status: 4 of 11 files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #139, opened 2021-09-28
Add SharedStorageTask for Copy{To,From}SharedStorageStep.

efoxepstein:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139)

### DIFF
--- a/build/repositories.bzl
+++ b/build/repositories.bzl
@@ -48,9 +48,9 @@ def panel_exchange_client_repositories():
 
     http_archive(
         name = "wfa_measurement_proto",
-        sha256 = "4045a4f05a37dc893096bcd0fe7392a23269560ca137b2164e39c3dd089c8cd6",
-        strip_prefix = "cross-media-measurement-api-0.7.0",
-        url = "https://github.com/world-federation-of-advertisers/cross-media-measurement-api/archive/v0.7.0.tar.gz",
+        sha256 = "ea943e080ffabe58e168c79706f395e4e2e0d3a5f2870546c714c2e41edcd915",
+        strip_prefix = "cross-media-measurement-api-0.7.1",
+        url = "https://github.com/world-federation-of-advertisers/cross-media-measurement-api/archive/v0.7.1.tar.gz",
     )
 
     http_archive(
--- a/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CustomIOExchangeTask.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CustomIOExchangeTask.kt
@@ -12,17 +12,21 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package org.wfanet.panelmatch.common.beam
+package org.wfanet.panelmatch.client.exchangetasks
 
-import org.wfanet.panelmatch.common.ShardedFileName
+import com.google.protobuf.ByteString
+import kotlinx.coroutines.flow.Flow
+import org.wfanet.panelmatch.client.storage.VerifiedStorageClient.VerifiedBlob
 
-internal class FileSpecBreakdown(fileSpecUri: String) {
-  val directoryUri: String
-  val shardedFileName: ShardedFileName
-
-  init {
-    val index = fileSpecUri.lastIndexOf('/')
-    directoryUri = fileSpecUri.substring(0, index + 1)
-    shardedFileName = ShardedFileName(fileSpecUri.substring(index + 1))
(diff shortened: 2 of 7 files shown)


stevenwarejones (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139)* status: 5 of 8 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/SharedStorageTask.kt, line 43 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139#-Mkn3E-52jS6budBzvxC:-Mkn3E-52jS6budBzvxD:blgql3f) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/db32c6f454a2c10a9d50aa0e8f77fdc56a3a6bec/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/SharedStorageTask.kt#L43)):*
> ```Kotlin
>       when (copyOptions.labelType) {
>         BLOB -> getBlobOutputs(inputBlob)
>         MANIFEST -> getManifestOutputs(inputBlob)
> ```

We should be reading and writing each one individually vs reading them all at once and then writing them. What if the total file size is 10gb?
___
*[src/main/kotlin/org/wfanet/panelmatch/common/ShardedFileName.kt, line 17 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139#-Mkn1WKZ3j4SMCcuN3tB:-Mkn1WKZ3j4SMCcuN3tC:bqbli96) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/db32c6f454a2c10a9d50aa0e8f77fdc56a3a6bec/src/main/kotlin/org/wfanet/panelmatch/common/ShardedFileName.kt#L17)):*
> ```Kotlin
> package org.wfanet.panelmatch.common
> 
> private val SPEC_REGEX = """^([^/]+)-(\*|\?+)-of-(\d+)$""".toRegex()
> ```

do you not need to delimit the first slash `^([^\/]+)`? 

Also the first capturing group is a little too broad for me. I'd prefer something more like
`^([\w\-.]+)-(\*|\?+)-of-(\d+)$`
___
*[src/main/kotlin/org/wfanet/panelmatch/common/ShardedFileName.kt, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139#-Mkn2jwqDHJa42-s00Rd:-Mkn2jwqDHJa42-s00Re:b1quq16) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/db32c6f454a2c10a9d50aa0e8f77fdc56a3a6bec/src/main/kotlin/org/wfanet/panelmatch/common/ShardedFileName.kt#L38)):*
> ```Kotlin
>     val matchResult = requireNotNull(SPEC_REGEX.matchEntire(spec)) { "Invalid spec: $spec" }
>     val matches = matchResult.groupValues
>     check(matches.size == 4)
> ```

i think if you go with the regex I suggest, you don't need to check the number of groupValues, right?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139)* status: 5 of 8 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/SharedStorageTask.kt, line 43 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139#-Mkn3E-52jS6budBzvxC:-MknEGM_CRkABQEu1hOT:bkoq1w6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/db32c6f454a2c10a9d50aa0e8f77fdc56a3a6bec/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/SharedStorageTask.kt#L43)):*

We should be reading and writing each one individually vs reading them all at once and then writing them. What if the total file size is 10gb?
</blockquote></details>

This doesn't actually read them -- `outputs` has type `Map<String, Flow<ByteString>>` and flows are lazy.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/139)* status: 4 of 8 files reviewed, all discussions resolved (waiting on @jonmolle and @stevenwarejones)

PR #157, opened 2021-10-09
Implement StorageClient-based I/O in Apache Beam.

efoxepstein:
This also changes some types around to reduce the number of expensive group-by-key operations when dealing with very large PCollections.

This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157)


efoxepstein (commented):
Advantages:

* The way signatures are handled is consistent with non-Beam outputs
* It's less code overall
* It's likely more efficient (since there's no need for another shuffle stage [group by key])

Disadvantages:

* It's less robust to errors
* It doesn't use atomic moves under the hood so it's possible for partial files to be written
* It requires each file's contents to fit into memory [but I guess this is fine if we just enforce that there are enough shards?]

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157#-:-MlotjbWCtGyzH-FR3rv:b-yefwh7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157)* status: 5 of 27 files reviewed, 3 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt, line 92 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157#-MltsW9IC0j93w80wRLZ:-MltsW9J2E5m-UFaelKR:btyd188) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ac977205690dbb28e43873d667027ffad8c9cc41/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt#L92)):*
> ```Kotlin
>     require(queryDecryptionKeysFileSpec.shardCount == outputs.queryIdAndPanelistKeyFileCount)
>     queryIdAndPanelistKeys
>       .keyBy { it.queryId.id % queryDecryptionKeysFileSpec.shardCount }
> ```

TODO: this doesn't work because we need to combine per key 
___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt, line 100 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157#-MltsDo93EZXVrokfjFz:-MltsDoAAoMFH4l8-7IR:bhjjicv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ac977205690dbb28e43873d667027ffad8c9cc41/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/BuildPrivateMembershipQueriesTask.kt#L100)):*
> ```Kotlin
>     require(encryptedQueriesFileSpec.shardCount == parameters.numShards)
>     encryptedQueryBundles
>       .mapKeys { it.id }
> ```

This will be problematic because we probably don't want like 10,000 files written.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExecutePrivateMembershipQueriesTask.kt, line 78 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157#-Mltsw4z-YStseG3WONt:-Mltsw4z-YStseG3WONu:b-mm0gf5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ac977205690dbb28e43873d667027ffad8c9cc41/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExecutePrivateMembershipQueriesTask.kt#L78)):*
> ```Kotlin
>     val encryptedResultsFileSpec =
>       ShardedFileName(outputs.encryptedQueryResultFileName, outputs.encryptedQueryResultFileCount)
>     results.mapKeys { it.id }.mapValues { it.toByteString() }.write(encryptedResultsFileSpec)
> ```

TODO: this can be problematic because there can reasonable been >10,000 shard ids and that's a LOT of files.

stevenwarejones (approved):
I'm good with this for now. 

> Reviewed 7 of 22 files at r3, 10 of 15 files at r4, 1 of 4 files at r5, 19 of 24 files at r7, 6 of 6 files at r8, 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/157#-:-MlwECF6DkFandLM0RxH:bfdiuik)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

PR #158, opened 2021-10-11
Initial Certificate Management class for Storage

jonmolle:
Adds a CertificateManager class that is in charge of retrieving and validating all certificates used in signing.

Also begins to implement the plan that private storage is _not_ signed, both due to redundancy and as a way to reduce interdependencies between tasks.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158)


jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158)* status: 11 of 14 files reviewed, 3 unresolved discussions (waiting on @duliomatos1, @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt, line 62 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158#-MlpsG6R9AzAFcD4BrhQ:-MlpuwH80bJo1MT6TooC:b-yuqbdn) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1fcb3305e3b073b4be42a1d70b8a67d9e958d924/src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt#L62)):*

why does this need to return a Pair, esp if one part of that pair was part of the input?
</blockquote></details>

Because: It used to not be part of the input and I didn't notice and remove it. Fixed.

SanjayVas (dismissed):
> Reviewed 2 of 14 files at r1, 1 of 1 files at r3, 1 of 3 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158)* status: all files reviewed, 4 unresolved discussions (waiting on @duliomatos1, @efoxepstein, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt, line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158#-MlmWCi85Ii_pCr-_Uir:-MlrGXYn5BokOs3YN7WI:b-unu2sg) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4b1959639a0aa46c2b4157b989c4fdf340dd5356/src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt#L60)):*

+1, I copied this from another repo, but I honestly have no idea whether this is placeholder code or not.
</blockquote></details>

Gah, looks like Reviewable ate my initial comment.

This is correct assuming that we're 1. Not using Tink primitives for digital signatures and 2. Are storing our private keys as DER-encoded PKCS#8 (which is what `PrivateKey.encoded` returns).

I've been also toying with how we want to do key storage for signing private keys. One thought is whether we should be storing the certificate along with the private key. It was my instinct to do so to parallel how Tink private keysets also include the public part of the key pair. The eventual goal would be to have the storage class return a handle that has a `sign` method, such that the actual `PrivateKey` is never exposed. See draft PR world-federation-of-advertisers/common-jvm#48
___
*[src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt, line 33 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158#-MlrEZ7k0v_xQRShEoBw:-MlrEZ7l-EwvNGr18wIB:b-pqoxy1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/d1197e0cba3901a9ea788cbfac30d9318730879a/src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt#L33)):*
> ```
> Grpc
> ```
This should probably be just a `CertificateManager` in some `v2alpha` package, as this uses the `v2alpha` servicesworld-federation-of-advertisers/common-jvm/pull/48

efoxepstein (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158)* status: all files reviewed, 3 unresolved discussions (waiting on @duliomatos1, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt, line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/158#-MlmWCi85Ii_pCr-_Uir:-MlrknY4A2Uxc5TfiDXs:bfljp3y) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/4b1959639a0aa46c2b4157b989c4fdf340dd5356/src/main/kotlin/org/wfanet/panelmatch/common/GrpcCertificateManager.kt#L60)):*

Gah, looks like Reviewable ate my initial comment.

This is correct assuming that we're 1. Not using Tink primitives for digital signatures and 2. Are storing our private keys as DER-encoded PKCS#8 (which is what `PrivateKey.encoded` returns).

I've been also toying with how we want to do key storage for signing private keys. One thought is whether we should be storing the certificate along with the private key. It was my instinct to do so to parallel how Tink private keysets also include the public part of the key pair. The eventual goal would be to have the storage class return a handle that has a `sign` method, such that the actual `PrivateKey` is never exposed. See draft PR world-federation-of-advertisers/common-jvm#48
</blockquote></details>

OK, let's merge this as is and revisit if once we make any new decisions for signing.

PR #172, opened 2021-10-15
Add isFirstExchange to IntersectAndValidate for bootstrapping.

efoxepstein:
This also changes the behavior to explicitly take as input the number of new items acceptable. This feels more realistic with how it will be used.


jonmolle (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/172)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @duliomatos1 and @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchange.kt, line 78 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/172#-MmJUlE8DRkoCbRMyxSD:-MmJUlE8DRkoCbRMyxSE:b-nkmlgf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/263f60cdf80a8ddceed11b7a77164e6aa915e784/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchange.kt#L78)):*
> ```Kotlin
>     require(step.stepCase == StepCase.INTERSECT_AND_VALIDATE_STEP)
>     val maxSize = step.intersectAndValidateStep.maxSize
>     // TODO: update v2alpha API proto to expose this instead
> ```

Nit: what's to keep a malicious kingdom from making this really high?

efoxepstein (commented):
Some validation is here: https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/main/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepValidatorImpl.kt#L68-L84

_Something_ will definitely need to surface whether or not this is the first Exchange. I'm thinking maybe `ExchangeStepValidator` should return something like `data class ExchangeContext(val workflow: ExchangeWorkflow, isFirstExchange: Boolean)` that is made available to `ExchangeTaskMapper`.

WDYT?

There's already a TODO in `ExchangeTaskMapperForJoinkeyExchange` that can cover this.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/172#-:-MmTZRZbBJ0Xmj8Ftc4T:b-vaicyo)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @duliomatos1)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/172)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @duliomatos1, @efoxepstein, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchange.kt, line 86 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/172#-MmTkFHp1xJOlO5akSYW:-MmZKhDWDxfUXUGbvFfY:b-zc4tor) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/6cbc3311a813f06509e14fd3ea69a6d2230df0c8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTaskMapperForJoinKeyExchange.kt#L86)):*

> so, this means we need a new recurring exchange every time we want to rotate the model provider's keys?

Nope. As long as the `previous-data` input has been properly re-keyed, MP key rotation is supported. This is something that we could do every single time, or we could do it in a separate ExchangeWorkflow. (Probably the simplest option is to just always rekey it.)

To re-key:

1. EDP generates a fresh DCE (deterministic commutative encryption, eg EcCommutativeCipher) key
2. EDP uses that to encrypt the old joinkeys
3. EDP sends to MP
4. MP removes their old DCE
5. MP adds their new DCE
6. MP sends to EDP
7. EDP removes their DCE

Note that MP can't cheat here because the data has the EDP's DCE. The only thing they can do is add/remove layers of encryption. Any other tampering will just wreck the resulting value in step 7.
</blockquote></details>

Are you okay with holding this PR in place until we can get this approved?

PR #194, opened 2021-11-01
Update ExchangeWorkflowDaemon to build storage selectors

jonmolle:
Previously we were hard-coding storage clients. Now that we're using storage selectors, we need to take in a bunch of secret maps (and custom Map maps of allowable factories) to build the storage system.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/194)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
@@ -20,6 +20,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import org.wfanet.measurement.common.logAndSuppressExceptionSuspend
 import org.wfanet.measurement.common.throttler.Throttler
+import org.wfanet.panelmatch.client.common.ExchangeContext
 import org.wfanet.panelmatch.client.common.Identity
 import org.wfanet.panelmatch.client.exchangetasks.ExchangeTaskMapper
 import org.wfanet.panelmatch.client.launcher.ApiClient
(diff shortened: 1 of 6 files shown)


efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/194)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @jonmolle)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt, line 73 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/194#-MnQex6o7HjJlTr4KzzF:-MnQex6o7HjJlTr4KzzG:b-cgszj6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1daf3165d1893b2277442fca13c3a1dfe9fa055f/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L73)):*
> ```Kotlin
> 
>   /** Map of supported shared storage platforms to their respective [StorageFactory] classes */
>   abstract val sharedStorageFactories:
> ```

I feel like we either need this or shared storage selector but not both here. Of we have this, shared storage selector need not be abstract, right?

jonmolle (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/194)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt, line 73 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/194#-MnQex6o7HjJlTr4KzzF:-Mnb-UY9D_3OfPk8s4lx:blv9fbk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1daf3165d1893b2277442fca13c3a1dfe9fa055f/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L73)):*

I feel like we either need this or shared storage selector but not both here. Of we have this, shared storage selector need not be abstract, right?
</blockquote></details>

For private you're right, but the shared gets the owner name from a flag (which I'm not entirely sure about, I just assumed the "id" was the resource name of whoever was running the binary). If that can be changed then they can be not-abstract. It also seemed a bit weird to have one abstract and not the other, but "a bit weird" is pretty weak design justification.

efoxepstein:
This PR has been folded into #198.

PR #197, opened 2021-11-03
Add Join key cryptor

stevenwarejones:
* Has the C++ for encrypting and hashing operate on JoinKeyAndId instead of just ByteString
* Adds hashing joinkeys as part of the decrypt step
* Simplifies inputs to CreateQueriesPipeline
* Simplifies inputs to DecryptQueryResultsPipeline

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197)


efoxepstein (changes requested):
> Reviewed 11 of 77 files at r5, 2 of 30 files at r6, 2 of 51 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197)* status: 15 of 111 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.cc, line 52 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni9R1rCp6wuzAaMcfb:-Mni9R1rCp6wuzAaMcfc:b-h5ajrk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.cc#L52)):*
> ```cpp
>   ASSIGN_OR_RETURN(SecretData key, generator.GenerateKey());
>   JoinKeyCryptorGenerateCipherKeyResponse response;
>   response.set_key(std::string(SecretDataAsStringView(key)).data());
> ```

Why is the `.data()` call necessary? There's a `set_key` that takes a `const std::string&`: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#proto3_string
___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h, line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni8v1w-8KPCnAdnfp-:-Mni8v1w-8KPCnAdnfp0:b-t8bduo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h#L15)):*
> ```objc
> // limitations under the License.
> 
> #ifndef SRC_MAIN_CC_WFA_PANELMATCH_CLIENT_JOINKEYEXCHANGE_JOIN_KEY_CRYPTOR_H_
> ```

I'm not sure if I agree with this change. There will be other things that need to use this functionality but aren't joinkeys. For example, the padding query mitigation relies on it for the results of the fake buckets. Why are you opposed to the generic name?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197)* status: 15 of 111 files reviewed, 2 unresolved discussions (waiting on @efoxepstein)

___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.cc, line 52 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni9R1rCp6wuzAaMcfb:-MnkW_TE7_eqAnBGBuEy:b-3r234k) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.cc#L52)):*

Why is the `.data()` call necessary? There's a `set_key` that takes a `const std::string&`: https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#proto3_string
</blockquote></details>

this was existing code I believe. I'll update it. Want to make sure we're on the same page on this PR though.
___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h, line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni8v1w-8KPCnAdnfp-:-MnkW1uZA9DQ_RoizNdz:bvfmjo2) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h#L15)):*

I'm not sure if I agree with this change. There will be other things that need to use this functionality but aren't joinkeys. For example, the padding query mitigation relies on it for the results of the fake buckets. Why are you opposed to the generic name? 
</blockquote></details>

I'm not a fan of passing `List<ByteString>`. It prevents doing the work out of order. I think its also cleaner whats happening. It still is using the common deterministic cryptor under the hood. It also makes the JoinKeyCryptor in kotlin less messy since it currently assumes that it sends a list of ByteString and gets them back in the same order and the it re-assigns the joinkeyid back in the same order. That, too, feels a little messy.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197)* status: 15 of 111 files reviewed, 2 unresolved discussions (waiting on @efoxepstein)

___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h, line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni8v1w-8KPCnAdnfp-:-MnkZ-N2BZAsZgHOLCbq:b-ieazxl) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h#L15)):*

I'm not a fan of passing `List<ByteString>`. It prevents doing the work out of order. I think its also cleaner whats happening. It still is using the common deterministic cryptor under the hood. It also makes the JoinKeyCryptor in kotlin less messy since it currently assumes that it sends a list of ByteString and gets them back in the same order and the it re-assigns the joinkeyid back in the same order. That, too, feels a little messy.
</blockquote></details>

i think it comes down to how generic do we want the data we serialize to be. I think I'm now leaning toward more meaningful types

efoxepstein (changes requested):
> Reviewed 1 of 77 files at r5, 1 of 51 files at r7.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197)* status: 17 of 111 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h, line 15 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/197#-Mni8v1w-8KPCnAdnfp-:-MnlKCOd1z6FNhmM2Afy:b-6h3l5t) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/e180af1bf05610e5a74cff9e176cdbf2b44d1ccc/src/main/cc/wfa/panelmatch/client/joinkeyexchange/join_key_cryptor.h#L15)):*

i think it comes down to how generic do we want the data we serialize to be. I think I'm now leaning toward more meaningful types
</blockquote></details>

I agree that `List<ByteString>` is maybe not the best decision. How about `Map<IdType, ByteString>` where `IdType` is an int or a string?

But I should note that in-order processing is _a thing_. Streaming gRPC guarantees message processing in the right order. "sed" and "awk" and all of the other unix utilities process lines in order.

PR #210, opened 2021-11-16
serialize interfaces potentially used in pipelines

stevenwarejones:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/eventpreprocessing/EventPreprocessor.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/eventpreprocessing/EventPreprocessor.kt
@@ -14,6 +14,7 @@
 
 package org.wfanet.panelmatch.client.eventpreprocessing
 
+import java.io.Serializable
 import org.wfanet.panelmatch.client.PreprocessEventsRequest
 import org.wfanet.panelmatch.client.PreprocessEventsResponse
 
@@ -23,7 +24,7 @@ import org.wfanet.panelmatch.client.PreprocessEventsResponse
  * method. The encrypted identifier is also used in an HKDF to create an AES key, which is used for
  * an AES encryption/decryption of event data.
  */
-interface EventPreprocessor {
+interface EventPreprocessor : Serializable {
   /** Preprocesses each of the events in [request] as described above */
   fun preprocess(request: PreprocessEventsRequest): PreprocessEventsResponse
 }
--- a/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt
@@ -15,11 +15,12 @@
 package org.wfanet.panelmatch.client.exchangetasks
 
 import com.google.protobuf.ByteString
+import java.io.Serializable
 import kotlinx.coroutines.flow.Flow
 import org.wfanet.measurement.storage.StorageClient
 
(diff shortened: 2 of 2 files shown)


efoxepstein (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210#-MoeIzdv3cSCra0BImvV:-MoeIzdw8GwJ_pIkvrLL:b3a2fpi) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/6e3cb3099ba4554eaafb18d4f1b15de470afcedc/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L23)):*
> ```Kotlin
> 
> /** Interface for ExchangeTask. */
> interface ExchangeTask : Serializable {
> ```

I think this is unlikely to actually work. You're very likely to run into issues with Beam claiming things are modified after they're serialized.

There's a reason we use `StorageFactory` instead of just a `StorageClient` in Beam pipelines. Basically, the factory is trivially serializable and then we don't rely on the `StorageClients` being serializable at all.

Some of the ExchangeTasks also close over things that definitely can't be serialized. For example, `InputTask` has a `StorageClient` and `GenerateExchangeCertificateTask` has a `CertificateManager`.

Should we discuss this more interactively on Slack/Zoom?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt, line 23 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210#-MoeIzdv3cSCra0BImvV:-MoeQhKL4Twjo73iRbVD:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/6e3cb3099ba4554eaafb18d4f1b15de470afcedc/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/ExchangeTask.kt#L23)):*

I think this is unlikely to actually work. You're very likely to run into issues with Beam claiming things are modified after they're serialized.

There's a reason we use `StorageFactory` instead of just a `StorageClient` in Beam pipelines. Basically, the factory is trivially serializable and then we don't rely on the `StorageClients` being serializable at all.

Some of the ExchangeTasks also close over things that definitely can't be serialized. For example, `InputTask` has a `StorageClient` and `GenerateExchangeCertificateTask` has a `CertificateManager`.

Should we discuss this more interactively on Slack/Zoom?
</blockquote></details>

Done.

efoxepstein (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/210)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #231, opened 2021-12-04
Add OpenSSL CertificateAuthority

efoxepstein:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt
@@ -15,18 +15,22 @@
 package org.wfanet.panelmatch.common.certificates
 
 import java.security.PrivateKey
-import java.security.PublicKey
 import java.security.cert.X509Certificate
 
-/** Creates X509Certificates signed by a protected, root private key. */
+/** Creates X509Certificates signed by a protected root private key. */
 interface CertificateAuthority {
+  data class Context(
+    val organization: String,
+    val commonName: String,
+    val hostname: String,
+    val validDays: Int,
+  )
 
   /**
-   * Creates a [PrivateKey] and corresponding [X509Certificate] signed by [rootPublicKey].
+   * Creates a [PrivateKey] and corresponding [X509Certificate] signed by a root private key and
+   * verifiable by the shared root public key.
    *
    * TODO(@efoxepstein): use SigningKeyHandle instead of a PrivateKey directly.
    */
-  suspend fun generateX509CertificateAndPrivateKey(
-    rootPublicKey: PublicKey
-  ): Pair<X509Certificate, PrivateKey>
+  suspend fun generateX509CertificateAndPrivateKey(): Pair<X509Certificate, PrivateKey>
 }
--- a/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/common/certificates/V2AlphaCertificateManager.kt
@@ -89,8 +89,7 @@ class V2AlphaCertificateManager(
       return existingKeys.certResourceName
     }
 
-    val rootKey = getRootCertificate(localName).publicKey
-    val (x509, privateKey) = certificateAuthority.generateX509CertificateAndPrivateKey(rootKey)
+    val (x509, privateKey) = certificateAuthority.generateX509CertificateAndPrivateKey()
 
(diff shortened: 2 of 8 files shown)


mpatel24 (approved):

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r1, 6 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt, line 30 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231#-MqG9opy8N0Opr7nXBXB:-MqG9opy8N0Opr7nXBXC:b4vo8dl) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de52212f833001e30dd1489af7359c24ad006d09/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt#L30)):*
> ```
> signed by a Root Certificate
> ```
nit: technically signing is done by the private key for a certificate. Usually rather than discussing what something is signed by, we document how it can be verified. e.g. "which can be verified using a root certificate."
___
*[src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt, line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231#-MqGCICc7K9Eff0kbeLI:-MqGCICc7K9Eff0kbeLJ:bwo5bp4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de52212f833001e30dd1489af7359c24ad006d09/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt#L34)):*
> ```Kotlin
>    * TODO(@efoxepstein): use SigningKeyHandle instead of a PrivateKey directly.
>    */
>   suspend fun generateX509CertificateAndPrivateKey(): Pair<X509Certificate, PrivateKey>
> ```

Shouldn't this take in the Context, as that's specific to a user certificate?

I would assume that a CA instance could be used to generate multiple user certificates.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt, line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231#-MqGCICc7K9Eff0kbeLI:-MqGH2V8CQTwdJ9A6wCc:bmfp3mo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de52212f833001e30dd1489af7359c24ad006d09/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt#L34)):*

Shouldn't this take in the Context, as that's specific to a user certificate?

I would assume that a CA instance could be used to generate multiple user certificates.
</blockquote></details>

The context can be set once at construction I think.

Once the context is set, we can generate arbitrarily many user certs without needing to change anything, right?

SanjayVas (approved):
> Reviewed 1 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @efoxepstein)

___
*[src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt, line 34 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/231#-MqGCICc7K9Eff0kbeLI:-MqKqNP-0QajhIG_jHTO:bhyodiv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de52212f833001e30dd1489af7359c24ad006d09/src/main/kotlin/org/wfanet/panelmatch/common/certificates/CertificateAuthority.kt#L34)):*

The context can be set once at construction I think.

Once the context is set, we can generate arbitrarily many user certs without needing to change anything, right?
</blockquote></details>

Oh, I suppose our use case doesn't really need different contexts for the same root cert.

PR #234, opened 2021-12-06
move GCS (and future AWS) dependencies into separate paths

jonmolle:
- Not all EDPs/MPs will want to run on any specific cloud
- This means that unused libraries can be ignored, rather than required
for generic installations
- DefaultStorageFactories renamed to ExampleStorageFactories as it is
only used by example code
- Example code moved into a deployment subdirectory, again, to more
easily isolate optional dependencies.

If we want a default, I recommend we either keep GCS and S3 lines commented out or to only include filesystem and reference the example.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/234)


efoxepstein (approved):
Awesome! Much tidier!

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/234#-:-MqGBpz11d6OFT2DNTc_:b4176ck)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones and @yunyeng)

yunyeng (approved):

PR #236, opened 2021-12-08
Fix logic bug in EventDataPreprocessor.

efoxepstein:
The bug is that the raw identifier is used to form the AES key used to encrypt each event. The correct value to use to form the AES key is the encrypted (but not hashed) identifier.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/236)


stevenwarejones (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/236)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @efoxepstein)

PR #253, opened 2021-12-23
Add maven deploy

stevenwarejones:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)


SanjayVas (changes requested):
> Reviewed 6 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[maven.bazelrc, line 4 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrc_ny58p-pNmpMn7ve:-Mrc_jIJ534VxQe3Dw6W:bi766g1) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/maven.bazelrc#L4)):*
> ```
> 
> # Since we don't bundle in libc, we should build on an older but still supported LTS Ubuntu.
> build:maven --host_platform=//build/platforms:ubuntu_18_04
> ```

This is an attestation of what environment you're building on. I don't think it'd be clear to the caller that by using the `maven` configuration, you're attesting that you're building on Ubuntu 18.04. Contrast this with say the `container` config, which is explicitly tied to a specific container image.

I would move this line out of this bazelrc and into publish.yml.
___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-MrcbUAL6NFjFYMnyQgm:b-kwyc2) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*
> ```YAML
> jobs:
>   publish:
>     # We run on an older LTS Ubuntu since libc is not bundled into the jar.
> ```

nit: Can libc ever be bundled in? My understanding was the whole reason libc on Linux causes headaches is that it must be specifically compiled to the distribution version, so you can't statically link or bundle it.

Therefore I think this description is inaccurate, and is really just meeting the host platform requirement for building our image targets.



_Code quote:_
```YAML
since libc is not bundled into the jar.
```
___
*[.github/workflows/publish.yml, line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcpJ_895WlIWw_8y2s:-MrcpJ_895WlIWw_8y2t:b-6vjehy) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L35)):*
> ```YAML
>       - name: Publish packages
>         run: |
>           bazel build \
> ```

You'll probably want to use `bazelisk` rather than `bazel` just to make sure that `.bazelversion` is properly handled.



_Code quote:_
```YAML
bazel
```
___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-Mrcjo4z2n84mOBZurvq:br93plc) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*
> ```YAML
>             --config=maven \
>             //src/main/kotlin/org/wfanet/panelmatch/client/deploy:DeployJar_deploy.jar
>           mvn deploy:deploy-file \
> ```

The normal way to deploy to a Maven repository with Bazel would be using a [`java_export`](https://github.com/bazelbuild/rules_jvm_external#publishing-to-external-repositories) target. Can you explain why you can't do that here?

e.g. defining said target and just `bazel run`ning it from here.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 87 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcgpnT8GHq1BqAW3bg:-MrcgpnT8GHq1BqAW3bh:b-k3352x) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L87)):*
> ```Starlark
> 
> java_binary(
>     name = "DeployJar",
> ```

nit: maybe a better name to avoid overloading the deploy JAR concept that already exists for rules_java? Any Java binary target has an implicit "<name>_deploy.jar" output (as you're already using) which is referred to as a deploy JAR. This means that the regular output of this target named "DeployJar" would not actually be a deploy JAR by that definition.



_Code quote:_
```Starlark
DeployJar
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-MrcqAlx-WxCel-Yi3v2:bowg7qt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*

nit: Can libc ever be bundled in? My understanding was the whole reason libc on Linux causes headaches is that it must be specifically compiled to the distribution version, so you can't statically link or bundle it.

Therefore I think this description is inaccurate, and is really just meeting the host platform requirement for building our image targets.
</blockquote></details>

we could use a musl distribution. This is what rust does. But, you are right that bundling libc is a pain. Not sure the best way to capture all that in a comment.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 6 of 7 files reviewed, 5 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-Mrd2X-JE_gSWUvtHm2g:b8ifojt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*

we could use a musl distribution. This is what rust does. But, you are right that bundling libc is a pain. Not sure the best way to capture all that in a comment.
</blockquote></details>

Switching to use musl would not be trivial, and I'm not sure we'd even want to. We have chosen a base image (distroless) that uses glibc, and we're just making sure we're meeting the build requirements for targeting that image.

That said, distroless Java is now based on the current Debian stable (bullseye) which has glibc 2.31. Therefore, I think we could switch to Ubuntu 20.04 (focal) here. We just need to make sure to correctly inform Bazel of this via `--host_platform`.

efoxepstein (approved):
> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 5 unresolved discussions (waiting on @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 3 of 7 files reviewed, 5 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[maven.bazelrc, line 4 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrc_ny58p-pNmpMn7ve:-MsaCLfl8R2VHVRId6pr:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/maven.bazelrc#L4)):*

This is an attestation of what environment you're building on. I don't think it'd be clear to the caller that by using the `maven` configuration, you're attesting that you're building on Ubuntu 18.04. Contrast this with say the `container` config, which is explicitly tied to a specific container image.

I would move this line out of this bazelrc and into publish.yml.
</blockquote></details>

Done.
___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-MsaCRoWDID5bPiO6ILr:b-cceej0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*

Switching to use musl would not be trivial, and I'm not sure we'd even want to. We have chosen a base image (distroless) that uses glibc, and we're just making sure we're meeting the build requirements for targeting that image.

That said, distroless Java is now based on the current Debian stable (bullseye) which has glibc 2.31. Therefore, I think we could switch to Ubuntu 20.04 (focal) here. We just need to make sure to correctly inform Bazel of this via `--host_platform`.
</blockquote></details>

I agree that switching to musl is non-trivial. I'm just saying there are methods to de-bundle libc. I re-wrote the comment. lmk if that fits better.

Additionally, I'd prefer to use an older version than 20.04. It will link and require a newer version of libc than some clients might support.
___
*[.github/workflows/publish.yml, line 35 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcpJ_895WlIWw_8y2s:-MsaF9au4mn8qORAMUNG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L35)):*

You'll probably want to use `bazelisk` rather than `bazel` just to make sure that `.bazelversion` is properly handled.
</blockquote></details>

Done.
___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-MsaITkF83PfCNanlFJP:buxxn2v) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

The normal way to deploy to a Maven repository with Bazel would be using a [`java_export`](https://github.com/bazelbuild/rules_jvm_external#publishing-to-external-repositories) target. Can you explain why you can't do that here?

e.g. defining said target and just `bazel run`ning it from here.
</blockquote></details>

How would bazel read in the maven_coordinates from github in this case? I guess we could generate a rule that creates it from the environment variable. I'm not sure that is so much better than this. Wdyt?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 87 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcgpnT8GHq1BqAW3bg:-MsaDDjyCxaRKBRXit06:b-swuwyx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L87)):*

nit: maybe a better name to avoid overloading the deploy JAR concept that already exists for rules_java? Any Java binary target has an implicit "<name>_deploy.jar" output (as you're already using) which is referred to as a deploy JAR. This means that the regular output of this target named "DeployJar" would not actually be a deploy JAR by that definition.
</blockquote></details>

how about MavenJar ?

efoxepstein (dismissed):
> Reviewed 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 5 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-Mt4iwqJ1GPzJDxCXzDm:bfgtte5) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*
The comment is okay now, but I think it still somewhat implies that bundling libc in the JAR is a desired solution.

> Additionally, I'd prefer to use an older version than 20.04. It will link and require a newer version of libc than some clients might support.

Fair point given that this is a library rather than a container image.
___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-Mt4nF8u9ifykpWFqbFJ:bbanmkj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

How would bazel read in the maven_coordinates from github in this case? I guess we could generate a rule that creates it from the environment variable. I'm not sure that is so much better than this. Wdyt?
</blockquote></details>

The issue is just getting the tag name from the release, yes?

The java_export rule implementation doesn't do Make variable expansion on `maven_coordinates`, but it does issue a Bash command so we should be able to use shell environment variables.

Something like

```
java_export(
  maven_coordinates = "org.wfanet:panel-exchange-client:${ARTIFACT_VERSION}",
  # ...
)
```

efoxepstein (commented):
> Reviewed 1 of 7 files at r4, 1 of 4 files at r5, 2 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 7 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 5 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[.github/workflows/publish.yml, line 21 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcbUAL6NFjFYMnyQgl:-MtIvDUjArR9JZrt8l3W:bnqd9z4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L21)):*

The comment is okay now, but I think it still somewhat implies that bundling libc in the JAR is a desired solution.

> Additionally, I'd prefer to use an older version than 20.04. It will link and require a newer version of libc than some clients might support.

Fair point given that this is a library rather than a container image.
</blockquote></details>

slightly updated the comment again
___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-MtJ8BH54CowBqOqRW0E:b412xen) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

The issue is just getting the tag name from the release, yes?

The java_export rule implementation doesn't do Make variable expansion on `maven_coordinates`, but it does issue a Bash command so we should be able to use shell environment variables.

Something like

```
java_export(
  maven_coordinates = "org.wfanet:panel-exchange-client:${ARTIFACT_VERSION}",
  # ...
)
```

</blockquote></details>

Will this work with Kotlin? https://bazelbuild.github.io/rules_kotlin/kotlin.html recommends using java_binary for release artifacts.

java_export like all rules that have deps, also requires src files but there aren't any java source files to include.
___
*[.github/workflows/publish.yml, line 44 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCmHTAk73MSuNcVll:-MtPCmHTAk73MSuNcVlm:b-q5rzun) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/.github/workflows/publish.yml#L44)):*
> ```YAML
>             --define "maven_repo=https://maven.pkg.github.com/world-federation-of-advertisers/panel-exchange-client" \
>             --define "maven_user=x-access-token" \
>             --define "maven_password=${{ secrets.GITHUB_TOKEN }}" \
> ```

will this show up in the logs? I don't see a different documented way to pass this to java export @sanjayvas 
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MtPCxf8A5q3rKf4B4vS:biz8ja6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*
> ```Starlark
>         ":provided-deps",
>     ],
>     maven_coordinates = "org.wfanet:panel-exchange-client-core:${{ github.event.release.tag_name }}",
> ```

@efoxepstein - are you okay with me making a release from this branch to test this out? I've tested it with an env variable but not with this env variable.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-MtPE0PK5warDGSjDe5C:bz5tzv6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

Will this work with Kotlin? https://bazelbuild.github.io/rules_kotlin/kotlin.html recommends using java_binary for release artifacts.

java_export like all rules that have deps, also requires src files but there aren't any java source files to include.
</blockquote></details>

i think we should be good here hopefully. take a look.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 4 of 11 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-MtPgVX3DodCMNDjnEKD:b-8z8p6m) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

i think we should be good here hopefully. take a look.
</blockquote></details>

Yes, it will work with Kotlin. See the minimal example in https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MtPhUxB7Z82BMm_I_cL:b-r36ht) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

@efoxepstein - are you okay with me making a release from this branch to test this out? I've tested it with an env variable but not with this env variable.
</blockquote></details>

I don't think this will work. See https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262#issuecomment-1012607681
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtPhCdT3_ZGFPh2Oy4Y:b-i2m06q) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*
> ```Starlark
> 
> java_binary(
>     name = "provided-deps",
> ```

What is this for? java_export should be able to find all of the Maven deps.

See the generated pom.xml from the minimal example: https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262#issuecomment-1012596904

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[.github/workflows/publish.yml, line 38 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-Mrcjo4z2n84mOBZurvp:-MtQEx-l3bPCAdCFSTnM:b-p0js9u) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/.github/workflows/publish.yml#L38)):*

Yes, it will work with Kotlin. See the minimal example in https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262
</blockquote></details>

yeah - added `tags = ["no-javadocs"],` based on your minimal example
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MtQFHXsARYYLy85zjJT:btgi97j) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

I don't think this will work. See https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262#issuecomment-1012607681
</blockquote></details>

It did work with an environment variable here: https://github.com/world-federation-of-advertisers/panel-exchange-client/packages/1161879

I just didn't test it by making an actual release.

Am I understanding you are you suggesting something else won't work?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtQFbrbDGyKYdGZTqiD:bdk9u95) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

What is this for? java_export should be able to find all of the Maven deps.

See the generated pom.xml from the minimal example: https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262#issuecomment-1012596904
</blockquote></details>

these are all the maven deps I don't want to be bundled in the jar.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MtQVPy15bj8GJfkUmrX:b2tz6lv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

It did work with an environment variable here: https://github.com/world-federation-of-advertisers/panel-exchange-client/packages/1161879

I just didn't test it by making an actual release.

Am I understanding you are you suggesting something else won't work?
</blockquote></details>

I'm actually not entirely sure how this is working, but pretty cool that it is!
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtQVVoH77SzIjs2H9In:b-kacl82) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

these are all the maven deps I don't want to be bundled in the jar.
</blockquote></details>

Anything that it already detects as a Maven dep should automatically get excluded from the JAR (this is actually pretty important, as we very much do *not* want to have to manually maintain a list of all transitive deps here).

The things we do want in the deploy_env are ones that exist on Maven but we pull in from another Bazel repo instead. Looking at the generated JAR on my end, the only such classes I'm seeing included are the JetBrains Kotlin and gRPC.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 4 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MtQYUVd1EPQZEaRXTg_:b-ua4vsz) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

I'm actually not entirely sure how this is working, but pretty cool that it is!
</blockquote></details>

Success with tagged releases! I had to change it to ${TAG_NAME} to get it to work: https://github.com/world-federation-of-advertisers/panel-exchange-client/packages/1161879 -- I'll analyze the jar on Tuesday and actually import it to our system to see if there are any issues.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtQYfH48HYASqNg3u3R:bur07e7) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Anything that it already detects as a Maven dep should automatically get excluded from the JAR (this is actually pretty important, as we very much do *not* want to have to manually maintain a list of all transitive deps here).

The things we do want in the deploy_env are ones that exist on Maven but we pull in from another Bazel repo instead. Looking at the generated JAR on my end, the only such classes I'm seeing included are the JetBrains Kotlin and gRPC.
</blockquote></details>

okay - I will test this out on Tuesday

efoxepstein (approved):
> Reviewed 2 of 4 files at r5, 1 of 4 files at r7, 1 of 1 files at r8, 3 of 3 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 7 files at r4, 2 of 4 files at r5, 1 of 3 files at r9, 2 of 3 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 44 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCmHTAk73MSuNcVll:-MtiNucr6riOZfqOIx9s:bnyyen8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/.github/workflows/publish.yml#L44)):*

will this show up in the logs? I don't see a different documented way to pass this to java export @sanjayvas 
</blockquote></details>

It's possible it ends up in GHA logs. I don't know if GHA automatically recognizes and obscures its own secrets, or if there's something we can do to ensure it isn't logged.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 87 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MrcgpnT8GHq1BqAW3bg:-MtiQ2Dx8cD6yEptcdbl:b1v9s5v) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/71dad771e488cbdfde7a39418ace20cfe72a97ac/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L87)):*

how about MavenJar ?
</blockquote></details>

Technically the java_export target doesn't correspond to a JAR, but to the process of exporting. Also UpperCamelCase tends to be used for JVM binaries and tests, so this likely should be snake_case instead.

Maybe just "maven"?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtiOUgG-IrvlGemgjjf:b-pqdufu) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

done
</blockquote></details>

What was the result? This list still seems rather large. To clarify, this should be the list of deps that have a corresponding Maven artifact, but which we're currently pulling from an alternate source.

Maybe split these up and indicate which Maven artifact they correspond to. My assumption is that you also need to specify the corresponding artifact in the `runtime_deps` of the `java_export` target.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 49 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtiP4NbCAxKDIgR68pJ:-MtiP4NbCAxKDIgR68pK:biun4o8) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15d78f5c23acf145537ccfa4eb021046582966ec/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L49)):*
> ```Starlark
>         "@tink_java//src/main/java/com/google/crypto/tink:binary_keyset_reader",
>         "@tink_java//src/main/java/com/google/crypto/tink:binary_keyset_writer",
>         "@tink_java//src/main/java/com/google/crypto/tink:hybrid_decrypt",
> ```

We pull in Tink Java by source at a non-release version. We're currently ahead of their Maven build, so these deps cannot be substituted with a Maven artifact.



_Code quote:_
```Starlark
        "@tink_java//src/main/java/com/google/crypto/tink:aead",
        "@tink_java//src/main/java/com/google/crypto/tink:binary_keyset_reader",
        "@tink_java//src/main/java/com/google/crypto/tink:binary_keyset_writer",
        "@tink_java//src/main/java/com/google/crypto/tink:hybrid_decrypt",
        "@tink_java//src/main/java/com/google/crypto/tink:hybrid_encrypt",
        "@tink_java//src/main/java/com/google/crypto/tink:kms_clients",
        "@tink_java//src/main/java/com/google/crypto/tink:registry_cluster",
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mti_fvX1WI6YwSM5O-3:b-vzshdm) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

What was the result? This list still seems rather large. To clarify, this should be the list of deps that have a corresponding Maven artifact, but which we're currently pulling from an alternate source.

Maybe split these up and indicate which Maven artifact they correspond to. My assumption is that you also need to specify the corresponding artifact in the `runtime_deps` of the `java_export` target.
</blockquote></details>

no - i'm doing something different. This is a list of bazel deps I don't want included in the final jar.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtjQ4vL02B40B2j3Rbx:bj3ufmo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

I don't understand the purpose of this, or rather why the set of Bazel deps not included in the final JAR wouldn't be exactly the set of non-Maven Bazel deps for which there are Maven artifacts. Everything that doesn't have a matching Maven artifact must be included in the JAR in order for it to work, otherwise you have unsatisfied deps.
</blockquote></details>

These are all dependencies that can be pulled in as Maven dependencies. For example, we shouldn't be bundling kotlin or kotlin coroutines in the jar. Its a bad practice to have a fat jar like that.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 49 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtiP4NbCAxKDIgR68pJ:-MtjQIzJB7zvO_ylxXBT:b-mtminj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15d78f5c23acf145537ccfa4eb021046582966ec/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L49)):*

We pull in Tink Java by source at a non-release version. We're currently ahead of their Maven build, so these deps cannot be substituted with a Maven artifact.
</blockquote></details>

Why can't we pull in a release version?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mtjd5fRAqCU00HItuju:b1nhouo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

These are all dependencies that can be pulled in as Maven dependencies. For example, we shouldn't be bundling kotlin or kotlin coroutines in the jar. Its a bad practice to have a fat jar like that.
</blockquote></details>

Okay, so then this list should indeed be the deps that we pull in from a source other than Maven that have a corresponding Maven artifact, as I mentioned. Everything we pull from Maven should be detected by java_export and shouldn't need to be listed here. My recommendation was to document which Maven artifact that each set of deps in this list corresponds to, and ensure that this list is minimal.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 49 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtiP4NbCAxKDIgR68pJ:-Mtjde2H5grR_ZD8o00L:b-u6n5j9) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15d78f5c23acf145537ccfa4eb021046582966ec/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L49)):*

Why can't we pull in a release version?
</blockquote></details>

We're currently relying on JWT functionality that has been marked as stable since June, but hasn't yet been put into a Maven release. The Tink team doesn't seem to be in any rush to cut a release. See https://github.com/google/tink/issues/541

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 4 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mtmp2AbAszPd55FPm_6:bxf1edw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Okay, so then this list should indeed be the deps that we pull in from a source other than Maven that have a corresponding Maven artifact, as I mentioned. Everything we pull from Maven should be detected by java_export and shouldn't need to be listed here. My recommendation was to document which Maven artifact that each set of deps in this list corresponds to, and ensure that this list is minimal.
</blockquote></details>

i listed out maven deps. Is this what you had in mind?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 49 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtiP4NbCAxKDIgR68pJ:-MtmlAoYD9MpuyC7VzsH:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/15d78f5c23acf145537ccfa4eb021046582966ec/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L49)):*

We're currently relying on JWT functionality that has been marked as stable since June, but hasn't yet been put into a Maven release. The Tink team doesn't seem to be in any rush to cut a release. See https://github.com/google/tink/issues/541
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtnR2_e0HWA4bCyFmU9:b-eu9ltx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

i listed out maven deps. Is this what you had in mind?
</blockquote></details>

Yes, though don't you need to add these as `artifact` `runtime_deps` to the `java_export` target? Meaning everything in this list should be something that *otherwise* ends up included into the Maven JAR and not properly referenced in the generated pom.xml list. If you have something in this list, it's getting excluded from the JAR and the corresponding artifact needs to be added to the java_export target so that it gets referenced in the generated pom.xml file.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtnZQVT-p6XjNACkxlv:b-d6tyq4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Yes, though don't you need to add these as `artifact` `runtime_deps` to the `java_export` target? Meaning everything in this list should be something that *otherwise* ends up included into the Maven JAR and not properly referenced in the generated pom.xml list. If you have something in this list, it's getting excluded from the JAR and the corresponding artifact needs to be added to the java_export target so that it gets referenced in the generated pom.xml file.
</blockquote></details>

i added them as deploy_env parameters. I looked into the jar and they aren't being included and we are able to build against it when we include those deps.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mtnbwjn-lIetiEA2AwN:bdw9lbf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

i added them as deploy_env parameters. I looked into the jar and they aren't being included and we are able to build against it when we include those deps.
</blockquote></details>

Sorry, perhaps I'm misunderstanding.

1. `java_export` automatically finds all Maven artifact deps and excludes them from the generated JAR, specifying the appropriate coordinates in the generated pom.xml.
2. There are some deps that we pull from another source, but which have a matching Maven artifact. For these deps only, we want to exclude them from the generated JAR (using `deploy_env`) and include the coordinates in the generated pom.xml instead (by specifying the Maven coordinates using the `artifact` macro as `runtime_deps` to the `java_export` target).

efoxepstein (approved):
> Reviewed 1 of 3 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 9 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtrvLpwATqJ1wstUJQv:b4mlxlt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Sorry, perhaps I'm misunderstanding.

1. `java_export` automatically finds all Maven artifact deps and excludes them from the generated JAR, specifying the appropriate coordinates in the generated pom.xml.
2. There are some deps that we pull from another source, but which have a matching Maven artifact. For these deps only, we want to exclude them from the generated JAR (using `deploy_env`) and include the coordinates in the generated pom.xml instead (by specifying the Maven coordinates using the `artifact` macro as `runtime_deps` to the `java_export` target).
</blockquote></details>

Thanks for the clarification. I really like that. Added it.

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mtsu_G-5QhNrv5gz7iW:b-co8zdw) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Thanks for the clarification. I really like that. Added it.
</blockquote></details>

Ugh, I did some experimenting and I may not have gotten this right. In particular, I'm not seeing the specified `artifact`s being included in the generated pom.xml file. From my casual reading of the `has_maven_deps` Bazel aspect used to collect the Maven coordinates, it seems like this should work.  I'm not entirely sure how to make sure specific artifacts get included without having to manually reproduce our dependency tree in the POM template.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mttp6cI1bTv6LQJQNuR:bq1w0sd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Ugh, I did some experimenting and I may not have gotten this right. In particular, I'm not seeing the specified `artifact`s being included in the generated pom.xml file. From my casual reading of the `has_maven_deps` Bazel aspect used to collect the Maven coordinates, it seems like this should work.  I'm not entirely sure how to make sure specific artifacts get included without having to manually reproduce our dependency tree in the POM template.
</blockquote></details>

What do you think if I add some TODOs? The jar as-is is usable.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mttq-bL-DZwbcjXQgzF:bq1uvvd) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

What do you think if I add some TODOs? The jar as-is is usable.
</blockquote></details>

If by usable do you mean only if you happen to include the missing deps?

If that's the case, go ahead and add a TODO to figure out how to specify additional Maven deps. Also make sure the deploy_env list is indeed minimal. e.g. when I tested, it looked like gson was already in pom.xml and not in the project JAR without having to specify it here.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-Mtu9mV55m5KYNCdwIbb:b-kveqeo) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

If by usable do you mean only if you happen to include the missing deps?

If that's the case, go ahead and add a TODO to figure out how to specify additional Maven deps. Also make sure the deploy_env list is indeed minimal. e.g. when I tested, it looked like gson was already in pom.xml and not in the project JAR without having to specify it here.
</blockquote></details>

Aha! I figured it out. I was testing with the kotlin-reflect and kotlin-stdlib-jdk7 artifacts, but these are ones where we explicitly override the Maven artifacts with Bazel targets.

I think the right way to do it is have a separate `maven_install` with a different name, and pass that name to the `artifact` macro.

I also did some digging re: minimal deploy_env to find out where some of our external library deps come from. One set comes from kt_jvm_grpc_library, which adds a dep to @com_github_grpc_grpc_kotlin//stub/src/main/java/io/grpc/kotlin:stub. This can be replaced by the io.grpc:grpc-kotlin-stub Maven artifact. Most of the others can likely be found by looking through which aliases in the //imports tree don't map to something in the `@maven` repo

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MtuP_Dd7MVnkKBD7KkJ:b-xf9eks) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Aha! I figured it out. I was testing with the kotlin-reflect and kotlin-stdlib-jdk7 artifacts, but these are ones where we explicitly override the Maven artifacts with Bazel targets.

I think the right way to do it is have a separate `maven_install` with a different name, and pass that name to the `artifact` macro.

I also did some digging re: minimal deploy_env to find out where some of our external library deps come from. One set comes from kt_jvm_grpc_library, which adds a dep to @com_github_grpc_grpc_kotlin//stub/src/main/java/io/grpc/kotlin:stub. This can be replaced by the io.grpc:grpc-kotlin-stub Maven artifact. Most of the others can likely be found by looking through which aliases in the //imports tree don't map to something in the `@maven` repo
</blockquote></details>

Updated the minimal working example, which has gotten the bulk of the unnecessary deps out of the maven project JAR.

https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 8 of 11 files reviewed, 3 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 36 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPhCdT3_ZGFPh2Oy4X:-MuDC8DC9aE2ksBX9lky:b-1zxu4o) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/3bb957edba62f41395d8530b38aad65f0cdb5817/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L36)):*

Updated the minimal working example, which has gotten the bulk of the unnecessary deps out of the maven project JAR.

https://github.com/world-federation-of-advertisers/panel-exchange-client/pull/262
</blockquote></details>

added your example in along with kotlin-test too

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r14, 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 4 unresolved discussions (waiting on @stevenwarejones)

___
*[WORKSPACE, line 72 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuDYTIRD1M7MRqKtApA:-MuDYTIRD1M7MRqKtApB:bdwb9bb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ca9b4539b49cf7d066b4ef707205f511bcd5835b/WORKSPACE#L72)):*
> ```Starlark
>     ],
>     excluded_artifacts = panel_exchange_client_maven_excluded_artifacts(),
>     fetch_sources = True,
> ```

Now that I think of it, I don't think we need fetch_sources for this. We were only fetching the sources for IntelliJ "go to definition" give a nice result, but these are just for Maven deployment.



_Code quote:_
```Starlark
    fetch_sources = True,
```
___
*[build/maven_deps.bzl, line 40 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuDZYun5W0vvRhohsXU:-MuDZYun5W0vvRhohsXV:bnyd47z) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ca9b4539b49cf7d066b4ef707205f511bcd5835b/build/maven_deps.bzl#L40)):*
> ```Starlark
>     artifact("org.jetbrains.kotlin:kotlin-reflect", "maven_export"),
>     artifact("org.jetbrains.kotlin:kotlin-stdlib-jdk7", "maven_export"),
>     artifact("org.jetbrains.kotlin:kotlin-test", "maven_export"),
> ```

This should only be referenced from tests, not listed as a dep for the main library.



_Code quote:_
```Starlark
    artifact("org.jetbrains.kotlin:kotlin-test", "maven_export"),
```

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas)

___
*[WORKSPACE, line 72 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuDYTIRD1M7MRqKtApA:-MuGdBKsBH7w-dSL-wJQ:b-kghtrs) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ca9b4539b49cf7d066b4ef707205f511bcd5835b/WORKSPACE#L72)):*

Now that I think of it, I don't think we need fetch_sources for this. We were only fetching the sources for IntelliJ "go to definition" give a nice result, but these are just for Maven deployment.
</blockquote></details>

good point. done.
___
*[.github/workflows/publish.yml, line 44 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCmHTAk73MSuNcVll:-MuGdNbFEL6KvulBs_58:b-6pp480) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/.github/workflows/publish.yml#L44)):*

It's possible it ends up in GHA logs. I don't know if GHA automatically recognizes and obscures its own secrets, or if there's something we can do to ensure it isn't logged.
</blockquote></details>

Redaction is pretty standard process for CI/CD

From https://docs.github.com/en/actions/security-guides/encrypted-secrets

```
Warning: GitHub automatically redacts secrets printed to the log, but you should avoid printing secrets to the log intentionally.
```
___
*[build/maven_deps.bzl, line 40 at r15](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuDZYun5W0vvRhohsXU:-MuGe9OAC9x5Aac0fohY:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ca9b4539b49cf7d066b4ef707205f511bcd5835b/build/maven_deps.bzl#L40)):*

This should only be referenced from tests, not listed as a dep for the main library.
</blockquote></details>

Done.

SanjayVas (changes requested):
> Reviewed 3 of 3 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[deploy/pom_template.xml, line 26 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuHVETcAS0g0PjToN9M:-MuHVETcAS0g0PjToN9N:b-t6imjc) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ece6d8ce8a47d6165802ad9433baa2ada67951cb/deploy/pom_template.xml#L26)):*
> ```XML
>         </repository>
>     </distributionManagement>
> ```

This template is missing the substitutions that make it useful, e.g. `{dependencies}`. Right now the target just copies this file verbatim.

See https://github.com/bazelbuild/rules_jvm_external/blob/master/docs/api.md#java_export

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuHVtP8DAWNxb8BuFv2:-MuHVtP9-iJdgAC1n-AE:b-3iq3s5) (no related file):*
Just to clarify for posterity what I mean when I refer to "looking at the generated JAR/pom file", I'm building the `//src/main/kotlin/org/wfanet/panelmatch/client/deploy:maven.publish` target and then looking at `bazel-bin/src/main/kotlin/org/wfanet/panelmatch/client/deploy/maven-pom.xml` and `bazel-bin/src/main/kotlin/org/wfanet/panelmatch/client/deploy/maven-project.jar` to see what's in them.

---

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @SanjayVas)

___
*[deploy/pom_template.xml, line 26 at r16](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MuHVETcAS0g0PjToN9M:-MuO6udHCOTSUatAa0Ua:bb7f2mf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/ece6d8ce8a47d6165802ad9433baa2ada67951cb/deploy/pom_template.xml#L26)):*

This template is missing the substitutions that make it useful, e.g. `{dependencies}`. Right now the target just copies this file verbatim.

See https://github.com/bazelbuild/rules_jvm_external/blob/master/docs/api.md#java_export
</blockquote></details>

done

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MuRzBLy-8pjWFZ6ADRY:b-lfyaop) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

Success with tagged releases! I had to change it to ${TAG_NAME} to get it to work: https://github.com/world-federation-of-advertisers/panel-exchange-client/packages/1161879 -- I'll analyze the jar on Tuesday and actually import it to our system to see if there are any issues.
</blockquote></details>

So now that we have the version being substituted into the pom template, I'm seeing `<version>${TAG_NAME}</version>` in the generated file. This is what I was referring to with not being sure how to substitute that from an environment variable, and likely needing to have a file with the current version number.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @SanjayVas)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MuSgz7j5AnGukNFu7_7:bjpv43h) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

So now that we have the version being substituted into the pom template, I'm seeing `<version>${TAG_NAME}</version>` in the generated file. This is what I was referring to with not being sure how to substitute that from an environment variable, and likely needing to have a file with the current version number.
</blockquote></details>

What do you suggest? I think we could just leave out version from the pom.xml for now?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-MuSvTb42rv1HU0zO8PB:b-lqq34d) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

What do you suggest? I think we could just leave out version from the pom.xml for now?
</blockquote></details>

Doesn't the pom.xml file get used for the Maven deployment process? Meaning isn't the version that specified in it significant?

My suggestion would be to have a version.bzl file with the current release version. This means that the release version comes from that file and not the GitHub release tag, and you bump the version in that file when about to cut a release. This is a pattern that I've seen in other repos. e.g. see [grpc-java's releasing doc](https://github.com/grpc/grpc-java/blob/master/RELEASING.md#common-variables) or [Tink's version.bzl](https://github.com/google/tink/blob/master/tink_version.bzl)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: 10 of 11 files reviewed, 2 unresolved discussions (waiting on @efoxepstein, @SanjayVas, and @stevenwarejones)

___
*[.github/workflows/publish.yml, line 23 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MukxmRvDFdTk5LGv_lO:-MukxmRvDFdTk5LGv_lP:b-egm4u2) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/d16dde9db257f6f554a6da5a487ce2527676d0c8/.github/workflows/publish.yml#L23)):*
> ```YAML
>     # Since it may not be desirable to bundle libc into the jar, we want to link to an older supported libc.
>     # Thus, we run on an older LTS Ubuntu.
>     runs-on: ubuntu-18.04
> ```

@sanjayvas  - it looks like this rule isn't actually running on Ubuntu 18.04. Is this coming from the container instead? We'd need to container to be on Ubuntu 18.04.

SanjayVas (commented):
> Reviewed 1 of 1 files at r17.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[.github/workflows/publish.yml, line 23 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MukxmRvDFdTk5LGv_lO:-MuleiuHE5SBuExyhaX2:b1qu1i4) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/d16dde9db257f6f554a6da5a487ce2527676d0c8/.github/workflows/publish.yml#L23)):*

@sanjayvas  - it looks like this rule isn't actually running on Ubuntu 18.04. Is this coming from the container instead? We'd need to container to be on Ubuntu 18.04.
</blockquote></details>

What do you mean? This `runs-on` is what tells GitHub Actions to use its Ubuntu 18.04 runner. The command is just calling Bazel via Bazelisk, so it isn't starting an inner container or anything. What sign are you seeing that this isn't actually running on Ubuntu 18.04?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[.github/workflows/publish.yml, line 36 at r17](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MulfjaD3GthEfgtplWa:-MulfjaD3GthEfgtplWb:b-jfiee0) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/d16dde9db257f6f554a6da5a487ce2527676d0c8/.github/workflows/publish.yml#L36)):*
> ```YAML
>       - name: Publish packages
>         run: |
>           bazelisk run \
> ```

Shouldn't you also be specifying `--config=maven` to these commands?

SanjayVas (approved):
> Reviewed 2 of 2 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel, line 92 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/253#-MtPCxf8A5q3rKf4B4vR:-Mv-npKM-EHY4Uuu3RtX:bdwvzys) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/1b5824f72ed4efb0236c6e79c250bb20c357b35b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/BUILD.bazel#L92)):*

I specified a release in the pox.xml. lmk if we're good to go.
</blockquote></details>

This is fine for now, but I think we should eventually have a clear version file at root that just holds the current release version. There may be other tasks that go along with updating this, such as updating the top-level README.md's instructions on how to depend on the library for both Bazel and Maven.

PR #265, opened 2022-01-19
Add status reporter

stevenwarejones:
Currently, the stepExecutor takes in an ApiClient to its constructor. This doesn't work for execution environments that do not have network access. Instead, we propose a statusReporter that writes out storage. One example has been written that uses a storage client but others could be written that use a sql database for example. 

I've also added a new parameter, jobId, that keeps track of the job across the stages. Its currently a UUID in the example daemon but could be passed in as a parameter if the daemon is used as a one-time worker.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/common/ExchangeContext.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/common/ExchangeContext.kt
@@ -21,6 +21,7 @@ import org.wfanet.panelmatch.common.ExchangeDateKey
 
 /** Contextual information about an ExchangeStepAttempt. */
 data class ExchangeContext(
+  val jobId: String,
   val attemptKey: ExchangeStepAttemptKey,
   val date: LocalDate,
   val workflow: ExchangeWorkflow,
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
@@ -24,7 +24,8 @@ import org.wfanet.panelmatch.client.common.Identity
 import org.wfanet.panelmatch.client.exchangetasks.ExchangeTaskMapper
 import org.wfanet.panelmatch.client.launcher.ApiClient
 import org.wfanet.panelmatch.client.launcher.CoroutineLauncher
-import org.wfanet.panelmatch.client.launcher.ExchangeStepLauncher
+import org.wfanet.panelmatch.client.launcher.ExchangeStepManager
+import org.wfanet.panelmatch.client.launcher.ExchangeStepReporter
 import org.wfanet.panelmatch.client.launcher.ExchangeStepValidatorImpl
 import org.wfanet.panelmatch.client.launcher.ExchangeTaskExecutor
 import org.wfanet.panelmatch.client.storage.PrivateStorageSelector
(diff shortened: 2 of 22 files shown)


efoxepstein (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @stevenwarejones and @zasadnyy)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoTZSd2yuc8lepV4pn:-MtoTZSd2yuc8lepV4po:b6o0cks) (no related file):*
Why does `ApiClient` need network access? You can implement an ApiClient that does anything, no?

---
___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt, line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoRtfmBPVBZKyltXMJ:-MtoRtfmBPVBZKyltXMK:b8y8yk9) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt#L101)):*
> ```Kotlin
> 
>   /** Generates of fetches a jobId for each task run. */
>   protected abstract val generateJobId: () -> String
> ```

abstract fun?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/ApiClient.kt, line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoSlz3AbnjMn7xKVTb:-MtoSlz3AbnjMn7xKVTc:bqhcns3) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ApiClient.kt#L36)):*
> ```Kotlin
> 
>   /** Reports ExchangeStepAttempt as complete (successfully or otherwise). */
>   suspend fun reportStepAttempt(
> ```

Why is this being renamed?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepReporter.kt, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoTBg37ooj_TJRYLkg:-MtoTBg37ooj_TJRYLkh:bw3g6ei) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepReporter.kt#L38)):*
> ```Kotlin
>   )
> 
>   suspend fun reportExecutionStatus(jobId: String)
> ```

What is this for? Why would you want to "store" but not "report"?
___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepManager.kt, line 37 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoSJbi9kEpPVSGCDlV:-MtoSJbi9kEpPVSGCDlW:b-jgjrkn) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepManager.kt#L37)):*
> ```Kotlin
>    */
>   suspend fun manageStep() {
>     val jobId = generateJobId()
> ```

Why not just use the step resource name as the job id?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265)* status: 0 of 23 files reviewed, 5 unresolved discussions (waiting on @efoxepstein, @stevenwarejones, and @zasadnyy)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoTZSd2yuc8lepV4pn:-MtpCIb_CCFGPWwtudMy:b-rx0mal) (no related file):*

Why does `ApiClient` need network access? You can implement an ApiClient that does anything, no?
</blockquote></details>

if it does anything, isn't it more like the Reporter I included below? Also, that, too, would require some refactoring to support a real grpc api client and a reporter api client

---
___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepReporter.kt, line 38 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoTBg37ooj_TJRYLkg:-MtpD2mE02bTuzfWFWGE:b2ehipt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepReporter.kt#L38)):*

What is this for? Why would you want to "store" but not "report"?
</blockquote></details>

I've tried to make it flexible to allow, if needed, each step to be run independently of the other steps.
1. Claim Step
2. Execute Step
3. Report Step
I think its helpful to think about what if each steps ran in its own coroutine. If each step ran in its own coroutine/process/worker/executor, each step needs to write to storage to communicate to the next step the output of that step.

However, this is only strictly required by the current code for the communication between the last two steps since Execute runs in a coroutine. But, if the claim step runs in a different environment than the execute, then it, too, should write its output to some storage/table so that the execute step in a different environment knows what to execute.
___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepManager.kt, line 37 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/265#-MtoSJbi9kEpPVSGCDlV:-MtpCdgBEqhWn-t2Cpgh:bwwunpt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/fabc4975a2d6ba230388c9d47fe417c031484924/src/main/kotlin/org/wfanet/panelmatch/client/launcher/ExchangeStepManager.kt#L37)):*

Why not just use the step resource name as the job id?
</blockquote></details>

step resource name isn't known until after the 1st step (claim step) of the process. I think an ID that is independent of the steps is more useful.

stevenwarejones:
Closing this for now. I'd like to think more about a better approach to do this.

PR #281, opened 2022-03-02
Move GCPKmsTink initialize into GoogleCloudExampleDaemonMain

yunyeng:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
@@ -17,8 +17,7 @@ package org.wfanet.panelmatch.client.deploy
 import java.time.Clock
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.launch
 import org.wfanet.measurement.common.logAndSuppressExceptionSuspend
 import org.wfanet.measurement.common.throttler.Throttler
 import org.wfanet.panelmatch.client.common.Identity
@@ -115,17 +114,13 @@ abstract class ExchangeWorkflowDaemon : Runnable {
         jobLauncher = launcher
       )
 
-    runBlocking { runDaemon(exchangeStepLauncher) }
+    scope.launch(CoroutineName("ExchangeWorkflowDaemon")) { runDaemon(exchangeStepLauncher) }
   }
 
   /** Runs [exchangeStepLauncher] in an infinite loop. */
   protected open suspend fun runDaemon(exchangeStepLauncher: ExchangeStepLauncher) {
-    withContext(CoroutineName("ExchangeWorkflowDaemon")) {
-      throttler.loopOnReady {
-        // All errors thrown inside the loop should be suppressed such that the daemon doesn't
-        // crash.
-        logAndSuppressExceptionSuspend { exchangeStepLauncher.findAndRunExchangeStep() }
-      }
+    throttler.loopOnReady {
+      logAndSuppressExceptionSuspend { exchangeStepLauncher.findAndRunExchangeStep() }
     }
   }
 }
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowDaemon.kt
@@ -18,6 +18,8 @@ import java.time.Clock
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withContext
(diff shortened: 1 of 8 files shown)


SanjayVas (approved):
> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/filesystem/FilesystemExampleDaemonMain.kt, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281#-MxBabL86WQGn9fAKuFB:-MxBabL86WQGn9fAKuFC:bl61sfh) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/824d29e7d1a7b073c930db3b7d0f9c7ca99a14bf/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/filesystem/FilesystemExampleDaemonMain.kt#L39)):*
> ```Kotlin
> 
>   @Option(names = ["--tink-key-uri"], description = ["Tink Key Uri"], required = true)
>   private lateinit var tinkKeyUri: String
> ```

nit: this flag can stay in the base class as protected, since all impls use Tink. The main difference is just which KMS client gets used.



_Code quote:_
```Kotlin
  @Option(names = ["--tink-key-uri"], description = ["Tink Key Uri"], required = true)
  private lateinit var tinkKeyUri: String
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @SanjayVas, @stevenwarejones, and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt, line 116 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281#-MxFq2XqEnUxfTzHEPPW:-MxFweU0-EaKZ7scPRvl:b-qj4651) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/866cbecd32ba5f5d68982754533fd1b84796bd96/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L116)):*

The target can use a hostname rather than an IP address. In fact, that's the more likely case.
</blockquote></details>

Sorry, looks like InetSocketAddress handles hostnames as well. Regardless, gRPC expects a string and not an `InetSocketAddress`, and can do it's own validation on the string.

See https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#forTarget-java.lang.String-

stevenwarejones (approved):
> Reviewed 2 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt, line 116 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281#-MxFq2XqEnUxfTzHEPPW:-MxGuagg82rfJpWxCTUE:bid53kx) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/866cbecd32ba5f5d68982754533fd1b84796bd96/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L116)):*

Sorry, looks like InetSocketAddress handles hostnames as well. Regardless, gRPC expects a string and not an `InetSocketAddress`, and can do it's own validation on the string.

See https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html#forTarget-java.lang.String-
</blockquote></details>

yeah - it handles hostnames as well. I'd like to keep it. Thx

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281)* status: all files reviewed, 1 unresolved discussion (waiting on @yunyeng)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt, line 116 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/281#-MxFq2XqEnUxfTzHEPPW:-MxH3U3P0dNNlIRnXmDC:b-4cysuf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/866cbecd32ba5f5d68982754533fd1b84796bd96/src/main/kotlin/org/wfanet/panelmatch/client/deploy/ExchangeWorkflowFlags.kt#L116)):*

yeah - it handles hostnames as well. I'd like to keep it. Thx
</blockquote></details>

1. That's inconsistent with how we specify service targets for every other binary in the system so far.
2. It doesn't provide any value, as the gRPC channel interface expects a string and does its own validation.
3. It's inaccurate, as technically gRPC supports more forms than even host:port.

PR #291, opened 2022-04-01
Use a SupervisorJob within CoroutineLauncher.

efoxepstein:
This prevents cancellations within the ExchangeStepExecutor from cancelling the CoroutineLauncher's scope. If this scope becomes cancelled then ExchangeStepLauncher.findAndRunExchangeStep will continue
to claim new work but immediately mark it as failed, as it tries to execute a coroutine in a cancelled scope.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/291)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt
@@ -17,7 +17,9 @@ package org.wfanet.panelmatch.client.launcher
 import kotlinx.coroutines.CoroutineName
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.plus
 import org.wfanet.measurement.api.v2alpha.ExchangeStep
 import org.wfanet.measurement.api.v2alpha.ExchangeStepAttemptKey
 import org.wfanet.panelmatch.client.launcher.ExchangeStepValidator.ValidatedExchangeStep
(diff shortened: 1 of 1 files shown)


yunyeng (approved):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/291)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle, @robinsons, and @stevenwarejones)

robinsons (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/291)* status: all files reviewed, 1 unresolved discussion (waiting on @efoxepstein, @jonmolle, and @stevenwarejones)

___
*[src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt, line 33 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/291#-MzXugLOCJD4eb5xcImB:-MzXugLOCJD4eb5xcImC:bxxv74l) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/2e33e0359b062fd0a1a65a0180d901cd1f626f29/src/main/kotlin/org/wfanet/panelmatch/client/launcher/CoroutineLauncher.kt#L33)):*
> ```Kotlin
> ) : JobLauncher {
>   override suspend fun execute(step: ValidatedExchangeStep, attemptKey: ExchangeStepAttemptKey) {
>     (scope + SupervisorJob()).launch(CoroutineName(attemptKey.toName())) {
> ```

nit: Won't this create more scopes than is necessary? It seems like a single scope with a supervisor job is all that is needed, and all steps would be launched as children of that.

Don't let this comment block merging. The current approach should solve the problem, which seems like the priority.

PR #330, opened 2022-05-05
Fix ReadAsSingletonPCollection: avoid Create.ofProvider.

efoxepstein:
Create.ofProvider will execute the provider before the workflow starts; the contents of the resulting PCollection are serialized into the workflow description itself. This causes failures for some large files under some runners.

The fix is to read the file as part of the pipeline in a ParDo.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/330)


stevenwarejones (approved):
> Reviewed 2 of 2 files at r1, 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/330)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jonmolle)

PR #339, opened 2022-05-18
Bundle at most 16 queries together

efoxepstein:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/339)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt
@@ -247,22 +247,29 @@ private class EncryptQueriesFn(
   @ProcessElement
   fun processElement(context: ProcessContext) {
     val unencryptedQueries = context.element().value.map { it.unencryptedQuery }
+    val keyPair = context.sideInput(keys)
 
-    val (encryptedQueries, time) =
-      withTime { cryptor.encryptQueries(unencryptedQueries, context.sideInput(keys)) }
+    for (chunk in unencryptedQueries.chunked(16)) {
+      context.output(encrypt(chunk, keyPair))
+    }
+  }
+
+  private fun encrypt(
+    unencryptedQueries: List<UnencryptedQuery>,
+    keyPair: AsymmetricKeyPair
+  ): EncryptedQueryBundle {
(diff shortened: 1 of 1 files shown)


stevenwarejones (commented):
what's the background on this why we think this is needed? Its running fine on my side without this change.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/339#-:-N2RVzMABWqio3b0OCU_:b-lhlkzu)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @robinsons)

robinsons (commented):
It's going to be sensitive to the selection of workflow parameters. With our current deployment, we are getting EncryptedQueryBundles that exceed 200MiB, and we suspect that Beam isn't too happy about that. By chunking we hope to cap the size of an EncryptedQueryBundle and improve the performance, while reducing the memory requirement for worker machines.

Obviously a similar outcome can be achieved by increasing the shard count, but we're trying to make it less sensitive to parameter choice.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/339#-:-N2T5ozj0g5aQ0ZmPpXW:beyz5li)* status: 0 of 1 files reviewed, all discussions resolved

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/339)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt` line 252 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/339#-N3dflZwBNsLSe3vEX6d:-N3dflZwBNsLSe3vEX6e:b-ndsa59) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/5cad76e88cef5e1647693a42e6ed95dedc6e1a97/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt#L252)):*
> ```Kotlin
>     val keyPair = context.sideInput(keys)
> 
>     for (chunk in unencryptedQueries.chunked(16)) {
> ```

wouldn't it be better to batch these based on size of the query vs just 16? This is inefficient for small query sizes.

PR #344, opened 2022-06-14
Use breakFusion() judiciously to reduce costs.

robinsons:
This function was initially introduced to encourage the job runner to scale the number of workers between stages, by materializing the output of one stage so that it could be processed by a different number of workers. However, materializing intermediate results and shuffling the data to different workers incurs a massive cost (at least, on GCP). It has been observed that this shuffle cost is the single largest line item for GCP-run exchanges, coming in at around 5x the cost of the next largest line item (which is vCPU usage).

In addition, the --dataflow-max-num-workers flag, which was added after breakFusion(), appears to accomplish the same goal of prompting worker scaling (in most cases), but without the associated shuffle cost.

For these reasons, this PR removes breakFusion() from almost all locations. The only locations where it remains are in CopyToSharedStorageBeamTask and CopyFromSharedStorageBeamTask, because it was observed that without breakFusion(), these do not scale workers (also, their total combined shuffle size is <50MiB).

This has been tested multiple times on GCP with an exchange of 1B synthetic EDP events and 800K MP queries. The impact on the MP alone was the following:

* Reduced per exchange shuffle usage from 46TiB to 5TiB.
* Reduced the sum of time in Beam tasks from 131m to 75m.
* Peak worker count was 25% less after removing breakFusion().

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/eventpreprocessing/PreprocessEvents.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/eventpreprocessing/PreprocessEvents.kt
@@ -24,7 +24,6 @@ import org.wfanet.panelmatch.client.privatemembership.DatabaseEntry
 import org.wfanet.panelmatch.client.privatemembership.databaseEntry
 import org.wfanet.panelmatch.client.privatemembership.encryptedEntry
 import org.wfanet.panelmatch.client.privatemembership.lookupKey
-import org.wfanet.panelmatch.common.beam.breakFusion
 import org.wfanet.panelmatch.common.beam.groupByKey
 import org.wfanet.panelmatch.common.beam.kvOf
 import org.wfanet.panelmatch.common.beam.map
@@ -98,6 +97,5 @@ class PreprocessEventsTransform(
           this.encryptedEntry = encryptedEntry { data = it.value }
         }
       }
-      .breakFusion("Break Fusion After Preprocess Events")
   }
 }
--- a/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/privatemembership/CreateQueries.kt
@@ -35,7 +35,6 @@ import org.wfanet.panelmatch.client.common.queryIdOf
 import org.wfanet.panelmatch.client.common.shardIdOf
 import org.wfanet.panelmatch.client.common.unencryptedQueryOf
 import org.wfanet.panelmatch.client.exchangetasks.JoinKeyIdentifier
-import org.wfanet.panelmatch.common.beam.breakFusion
 import org.wfanet.panelmatch.common.beam.createSequence
 import org.wfanet.panelmatch.common.beam.filter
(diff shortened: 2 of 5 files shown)


stevenwarejones (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344)* status: all files reviewed, 1 unresolved discussion (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteShardedData.kt` line 87 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344#-N4gXFI18tU1pUukD7Qj:-N4gXFI18tU1pUukD7Qk:b-vjaxlf) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de35ad89f828ac4bd9ddc5232750f270802997e5/src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteShardedData.kt#L87)):*
> ```Kotlin
>         .and(missingFiles)
>         .flatten("Flatten groupedData+missingFiles")
>         .breakFusion("Break Fusion Before WriteFilesFn")
> ```

performance so far looks good but not sure about removing this breakFusion

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344)* status: all files reviewed, 1 unresolved discussion (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteShardedData.kt` line 87 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344#-N4gXFI18tU1pUukD7Qj:-N4hur9x8cFJxtmfX00z:b10z5al) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/de35ad89f828ac4bd9ddc5232750f270802997e5/src/main/kotlin/org/wfanet/panelmatch/common/beam/WriteShardedData.kt#L87)):*
<details><summary><i>Previously, robinsons wrote…</i></summary><blockquote>

I went ahead and tried this out, but it almost doubled the amount of shuffle data with no improvement in runtime. Testing was on GCP with 800K queries.

Based on my understanding, the aspect of breakFusion() that actually prevents fusion of stages is the combination of KeyBy+GroupByKey. We're already doing this combo about 10-15 lines above, so we're effectively breaking fusion even without the explicit call.
</blockquote></details>

i think the concern that I haven't tested is that if there are a large number of files 100k+, this will go slow, since the missing files creation happens after the keyby+groupby. Let me test that out today.

robinsons (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/344)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @robinsons)

PR #350, opened 2022-07-20
add initial support for passing aws creds to Dataflow

stevenwarejones:
Tested using a custom local Daemon

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)

### DIFF
--- a/build/maven_deps.bzl
+++ b/build/maven_deps.bzl
@@ -64,11 +64,13 @@ _ARTIFACTS = {
     "org.apache.beam:beam-sdks-java-extensions-google-cloud-platform-core": _BEAM_VERSION,
     "org.apache.beam:beam-sdks-java-extensions-protobuf": _BEAM_VERSION,
     "org.apache.beam:beam-sdks-java-io-google-cloud-platform": _BEAM_VERSION,
+    "org.apache.beam:beam-sdks-java-io-amazon-web-services": _BEAM_VERSION,
     "org.apache.beam:beam-vendor-guava-26_0-jre": "0.1",
     "org.hamcrest:hamcrest": "2.2",
     "org.slf4j:slf4j-simple": "1.7.32",
     "software.amazon.awssdk:utils": "2.17.100",
     "software.amazon.awssdk:sts": "2.17.100",
+    "software.amazon.awssdk:auth": "2.17.100",
 }
 
 _EXCLUDED_ARTIFACTS = [
--- a/imports/java/software/amazon/awssdk/auth/BUILD.bazel
+++ b/imports/java/software/amazon/awssdk/auth/BUILD.bazel
@@ -1,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+alias(
+    name = "auth",
+    actual = "@maven//:software_amazon_awssdk_auth",
+)
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/BUILD.bazel
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/BUILD.bazel
@@ -9,6 +9,7 @@ kt_jvm_library(
     srcs = glob(["*.kt"]),
     deps = [
         "//imports/java/org/apache/beam/runners:google_cloud_dataflow",
+        "//imports/java/software/amazon/awssdk/auth",
         "//src/main/kotlin/org/wfanet/panelmatch/client/deploy",
         "//src/main/kotlin/org/wfanet/panelmatch/client/deploy/example",
         "//src/main/kotlin/org/wfanet/panelmatch/client/storage",
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt
@@ -170,6 +170,7 @@ private class GoogleCloudExampleDaemon : ExampleDaemon() {
       val awsCredentials =
         DefaultCredentialsProvider.create().resolveCredentials() as AwsSessionCredentials
       // TODO: Encrypt using KMS or store in Secrets
+      // Think about moving this logic to a CredentialsProvider
       baseOptions.apply {
         awsAccessKey = awsCredentials.accessKeyId()
         awsSecretAccessKey = awsCredentials.secretAccessKey()
--- a/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt
@@ -18,7 +18,6 @@ import com.google.crypto.tink.integration.gcpkms.GcpKmsClient
 import java.util.Optional
 import kotlin.properties.Delegates
 import org.apache.beam.runners.dataflow.DataflowRunner
-import org.apache.beam.runners.dataflow.options.DataflowPipelineOptions
 import org.apache.beam.runners.dataflow.options.DataflowWorkerLoggingOptions
 import org.apache.beam.sdk.options.PipelineOptions
 import org.apache.beam.sdk.options.PipelineOptionsFactory
@@ -32,6 +31,7 @@ import org.wfanet.panelmatch.client.deploy.CertificateAuthorityFlags
 import org.wfanet.panelmatch.client.deploy.DaemonStorageClientDefaults
 import org.wfanet.panelmatch.client.deploy.example.ExampleDaemon
 import org.wfanet.panelmatch.client.storage.StorageDetailsProvider
+import org.wfanet.panelmatch.common.beam.BeamOptions
 import org.wfanet.panelmatch.common.certificates.gcloud.CertificateAuthority
 import org.wfanet.panelmatch.common.certificates.gcloud.PrivateCaClient
 import org.wfanet.panelmatch.common.secrets.MutableSecretMap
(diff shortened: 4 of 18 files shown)


robinsons (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyFromSharedStorageBeamTask.kt` line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7SrlFG2tO5qMgvC6WN:-N7SrlFG2tO5qMgvC6WO:bhz3tw6) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyFromSharedStorageBeamTask.kt#L57)):*
> ```Kotlin
>     pipeline.apply("Generate Shard Names", Create.of(shardedFileName.fileNames.asIterable()))
>   val writeFilesDoFn =
>     object : DoFn<String, String>() {
> ```

Could we make a dedicated class for this instead of an anonymous object?
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyToSharedStorageBeamTask.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7Srvtj4AxKFY4K1gVj:-N7Srvtj4AxKFY4K1gVk:brhumzt) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyToSharedStorageBeamTask.kt#L51)):*
> ```Kotlin
>   }
>   val readFilesDoFn =
>     object : DoFn<String, String>() {
> ```

Same here.

jonmolle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: 11 of 15 files reviewed, 3 unresolved discussions (waiting on @jonmolle, @robinsons, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt` line 170 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7WtnmWDmQzBvUzbMuW:-N7WtnmWDmQzBvUzbMuX:b-qo1xnj) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt#L170)):*
> ```Kotlin
>       // aws-sdk-java-v2 casts responses to AwsSessionCredentials if its assumed you need a
>       // sessionToken
>       val awsCredentials =
> ```

Does this assume you're going to be using the same AWS Credentials for all S3 connections across all exchanges?



_Code quote:_
```Kotlin
        DefaultCredentialsProvider.create().resolveCredentials() as AwsSessionCredentials
```

jonmolle (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: 11 of 19 files reviewed, 3 unresolved discussions (waiting on @jonmolle, @robinsons, and @stevenwarejones)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt` line 170 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7WtnmWDmQzBvUzbMuW:-N7X-06LEQetoulVYgpf:bfv7xqk) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt#L170)):*

Does this assume you're going to be using the same AWS Credentials for all S3 connections across all exchanges?
</blockquote></details>

Sorry, to follow this up, I think the paradigm is to keep all factory options within the storage details of the appropriate storageclient. I realize that we can't just have a session token hard-coded somewhere. I ask about how the credentials here are generated in part because if possible it would be cleaner to move this into the s3 storage details if possible. From what I can gather from this PR, that would involve some sort of check (either implicitly through checking existence of aws credentials or explicitly stating) whether we expect to get AWS credentials from the environment or the details object. That seems like it would allow us to support multiple different credentials per-partner, though, which I'm not sure this does. Am I off-base here? I'm also not against making this change is a downstream PR if that TODO on the line below this is referring to this change.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: 10 of 19 files reviewed, 3 unresolved discussions (waiting on @jonmolle and @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt` line 170 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7WtnmWDmQzBvUzbMuW:-N7b-G3M3JuxgV-QbEej:b-iht1qv) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt#L170)):*

Sorry, to follow this up, I think the paradigm is to keep all factory options within the storage details of the appropriate storageclient. I realize that we can't just have a session token hard-coded somewhere. I ask about how the credentials here are generated in part because if possible it would be cleaner to move this into the s3 storage details if possible. From what I can gather from this PR, that would involve some sort of check (either implicitly through checking existence of aws credentials or explicitly stating) whether we expect to get AWS credentials from the environment or the details object. That seems like it would allow us to support multiple different credentials per-partner, though, which I'm not sure this does. Am I off-base here? I'm also not against making this change is a downstream PR if that TODO on the line below this is referring to this change.  
</blockquote></details>

A couple clarifying points:
1. These are short-lived credentials that typically only last one hour
2. They are generated in the ExchangeTaskMapper when the task is generated on the client side
3. I'm not sure about the best way to pull this into the storageFactory config. The storageFactory config is set one time and used over and over again. These creds change all the time.

I've added a TODO to think about this a little more
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyFromSharedStorageBeamTask.kt` line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7SrlFG2tO5qMgvC6WN:-N7b3oN_4Abu1D3S6UMV:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyFromSharedStorageBeamTask.kt#L57)):*
<details><summary><i>Previously, robinsons wrote…</i></summary><blockquote>

Could we make a dedicated class for this instead of an anonymous object?
</blockquote></details>

Done.
___
*[`src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyToSharedStorageBeamTask.kt` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7Srvtj4AxKFY4K1gVj:-N7b3olZC4KQ49v-JUyT:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/exchangetasks/CopyToSharedStorageBeamTask.kt#L51)):*
<details><summary><i>Previously, robinsons wrote…</i></summary><blockquote>

Same here.
</blockquote></details>

Done.

jonmolle (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: 10 of 19 files reviewed, 2 unresolved discussions (waiting on @robinsons)

___
*[`src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt` line 170 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350#-N7WtnmWDmQzBvUzbMuW:-N7b4vBODiZt_weWQaPi:b-2n0hdb) ([raw file](https://github.com/world-federation-of-advertisers/panel-exchange-client/blob/b1c70d12ce37875e0fb74c1e47f514c534dde5c8/src/main/kotlin/org/wfanet/panelmatch/client/deploy/example/gcloud/GoogleCloudExampleDaemon.kt#L170)):*

A couple clarifying points:
1. These are short-lived credentials that typically only last one hour
2. They are generated in the ExchangeTaskMapper when the task is generated on the client side
3. I'm not sure about the best way to pull this into the storageFactory config. The storageFactory config is set one time and used over and over again. These creds change all the time.

I've added a TODO to think about this a little more
</blockquote></details>

That's good enough for me at least for now (IE: for a PoC). I think some of this may be a non-issue; specifically since you can grant a single service account access to multiple buckets, having a single aws account associated with the daemon is probably okay. I was more thinking about if it might be doable to actually generate these session credentials within the factory (moving the long-lived keys that are pulled here to storageDetails fields). If that's infeasible this seems okay, though.

robinsons (approved):
> Reviewed 2 of 4 files at r2, 4 of 4 files at r3, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/350)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #357, opened 2022-10-07
Fix algorithm to use one supported by AWS PrivateCAs (IE: not ECDSA)

jonmolle:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/357)

### DIFF
--- a/src/main/kotlin/org/wfanet/panelmatch/client/storage/StorageDetailsProvider.kt
+++ b/src/main/kotlin/org/wfanet/panelmatch/client/storage/StorageDetailsProvider.kt
@@ -21,12 +21,17 @@ import org.wfanet.panelmatch.common.secrets.SecretMap
  * Wraps a [SecretMap] to provide [StorageDetails].
  *
  * @param secretMap map from recurring exchange ids to serialized [StorageDetails] protos.
+ * @param detailsType is a string used only for debugging strings.
  */
-class StorageDetailsProvider(private val secretMap: MutableSecretMap) {
+class StorageDetailsProvider(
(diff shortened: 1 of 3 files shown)


sandip80:
Why are we moving to RSA? AWS private CA supports ECDSA right: https://docs.aws.amazon.com/privateca/latest/userguide/supported-algorithms.html ?

bdomen-ggl (commented):
I'll try to elaborate the issue we ran into.

Yes, specifically the AWS private CA in isolation does support ECDSA. Meaning, we can create a PCA using either ECDSA or RSA.

However, the problem is that when we try to make a request for the private key, we can't do so if the cryptographic algorithm is ECDSA. AWS only lets us request the private key if the cryptographic algorithm is RSA.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/357#-:-NE1G1vMAIb500RoJh8o:biu5ze0)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sandip80)

jonmolle (commented):
I'm trying to find the exact docstring, but basically you can make a Private CA backed with ECDSA but you can't issue certificates with private keys from it. Try it yourself, you won't see any private ECDSA-backed CAs as an option in the UI (and that's the same in the sdk/cli)

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/357#-:-NEK-xig9p5tGWznKehz:bij1xum)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @sandip80)

PR #365, opened 2023-05-08
Update panel exchange client to use Virtual People Common 0.2.3

jonmolle:
This gives access to more fields in DataProviderEvent, but otherwise should be a no-op from what I've been able to test.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/365)

PR #366, opened 2023-05-08
Hardcode AWS default algorithm as RSA.

jonmolle:
It is currently unset, but this removes a pitfall in setup.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/panel-exchange-client/366)

